PR #349, opened 2021-10-11
Update user balances in real-time

lazarmitic:


packages/tempus-client/src/services/ERC20TokenService.ts
```diff
@@ -23,7 +23,7 @@ type ERC20TokenServiceParameters = {
 };
 
 class ERC20TokenService {
-  private contract: ERC20 | null = null;
+  public contract: ERC20 | null = null;
```
danieledamiani, line 26: is this intentional?

packages/tempus-client/src/services/ERC20TokenService.ts
```diff
@@ -127,13 +127,22 @@ class ERC20TokenService {
     }
   }
 
-  async onTransfer(from: string | null, to: string | null, listener: TransferEventListener) {
+  onTransfer(from: string | null, to: string | null, listener: TransferEventListener) {
```
danieledamiani, line 130: missing return type

packages/tempus-client/src/services/ERC20TokenService.ts
```diff
@@ -127,13 +127,22 @@ class ERC20TokenService {
...
     if (!this.contract) {
       console.error('ERC20TokenService - approve() - Attempted to use ERC20TokenService before initializing it!');
       return Promise.reject();
     }
 
     this.contract.on(this.contract.filters.Transfer(from, to), listener);
   }
+
+  offTransfer(from: string | null, to: string | null, listener: TransferEventListener) {
+    if (!this.contract) {
```
danieledamiani, line 140: same as line 130

packages/tempus-client/src/services/ERC20TokenService.ts
lazarmitic: Yes, I'm accessing contract attribute from `getERC20TokenService.ts`. Problem is, every time we call `getERC20TokenService()` with signer, init() is called and new Contract instance is created. Because of this, all subscriptions we created are lost every time we call `getERC20TokenService()`.

I added additional check that checks if current contract already has same signer, is it does, we skip calling `init()`.

In total we can have two providers, Alchemy and MetaMask, I was even thinking of creating two separate instances of ERC20TokenService inside `getERC20TokenService`, because if I subscribe to the event using Alchemy provider, and try to react to that event using MetaMask provider, it's not gonna work.

packages/tempus-client/src/services/ERC20TokenService.ts
```diff
@@ -127,13 +127,22 @@ class ERC20TokenService {
     }
   }
 
-  async onTransfer(from: string | null, to: string | null, listener: TransferEventListener) {
+  onTransfer(from: string | null, to: string | null, listener: TransferEventListener) {
     if (!this.contract) {
```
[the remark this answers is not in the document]
lazarmitic: Fixed.

packages/tempus-client/src/services/ERC20TokenService.ts
lazarmitic: Fixed.

PR #422, opened 2021-10-21
Add dashboard TVL periodic updates

lazarmitic:


packages/tempus-client/src/adapters/DashboardDataAdapter.ts
```diff
@@ -51,6 +53,22 @@ export default class DashboardDataAdapter {
     return [...parentRows, ...childRows];
   }
 
+  public getTempusPoolTVL(tempusPool: string, backingTokenTicker: Ticker): Observable<BigNumber | null> {
+    if (!this.statisticsService) {
+      return of(null);
```
danieledamiani, line 58: if you prefer, you can `return throwError(() => new Error());` and the return type would be `Observable<BigNumber>`

lazarmitic: I intentionally return null, then down the line, if TVL is null I show loading circle in the dashboard, letting use know that it's still loading.

lazarmitic: Loading circle inside TVL cell, instead of TVL value.

danieledamiani (approved):

PR #434, opened 2021-11-01
Hide negative yield pools if user has no balance in them

danieledamiani:


packages/tempus-client/src/components/dashboard/dashboard.tsx
```diff
@@ -186,9 +188,34 @@ const Dashboard: FC<DashboardProps> = ({ hidden, userWalletAddress, rows, onRowA
...
+      const poolMap: { [address: string]: any } = {};
+      poolData.forEach(pool => {
+        poolMap[pool.address] = pool;
+      });
+
+      let rowsToDisplay = rows.filter((row: DashboardRow) => {
+        const pool = poolMap[row.id];
+
+        if (pool) {
+          if (pool?.variableAPR > 0) {
```
lazarmitic, line 205: I think we need to check if underlying protocol APR is less then zero. Even if underlying protocol is negative, final APR can be positive because we add swap fees on top of underlying APR.

danieledamiani: fixed

lazarmitic (approved):

PR #442, opened 2021-11-02
Add dashboard

lazarmitic:


packages/tempus-client_v2/src/components/main/Main.tsx
```diff
@@ -1,36 +1,11 @@
...
-          <Term language={language} />
-          <CurrentPosition language={language} />
-        </div>
-        <div className="tc__dashboard__row">
-          {selectedView === 'Deposit' && <Deposit />}
-
-          {showExtraInfo && <ProfitLoss language={language} />}
-        </div>
-      </div>
+      <DashboardManager />
```
danieledamiani, line 8: how to show Transaction view? I couldn't find in `dashboard.tsx` either

danieledamiani (changes requested):
a comment

packages/tempus-client_v2/src/components/main/Main.tsx
lazarmitic: I didn't add that part, but it will be similar to first version - dashboard.tsx has a callback prop onRowActionClick() that will set selected pool address in PoolDataContext -> selectedPool. If there is a selected pool, main.tsx will show pool view instead of the dashboard. I wanted to add this in next PR so you don't have to wait to much for my changes to start working on other things.

lazarmitic: https://github.com/tempus-finance/tempus-app/blob/bef2f346058a8ff11147d99e954739580314cd9d/packages/tempus-client_v2/src/components/dashboard/dashboard.tsx dashboard.tsx

danieledamiani (approved):

PR #480, opened 2021-11-05
Lazar/v2 fixes

lazarmitic:
Fixed mint operation.
Added Fees tooltip inside Pool card.
Removed settings link.
Fixed wallet connect button UI.
Withdraw is hidden if user does not have any balance in the pool.
Swap is hidden if user does not have any principals and yields.
Added min-width to token selector component.

### DIFF
--- a/packages/tempus-client_v2/src/components/pool/feesTooltip/feesTooltip.tsx
+++ b/packages/tempus-client_v2/src/components/pool/feesTooltip/feesTooltip.tsx
@@ -1,0 +1,92 @@
+import { BigNumber } from '@ethersproject/bignumber';
+import { ethers } from 'ethers';
+import { useContext, useEffect, useMemo, useState } from 'react';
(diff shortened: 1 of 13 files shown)


packages/tempus-client_v2/src/components/pool/feesTooltip/feesTooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+    }
+    return NumberUtils.formatPercentage(Number(ethers.utils.formatEther(poolFees[3])));
+  }, [poolFees]);
+
+  return (
+    <div className="tc__feesTooltip">
+      <Typography variant="card-title">Fees</Typography>
+      <div className="tc__feesTooltip-row">
+        <Typography variant="card-body-text" color="title">
+          Deposit
```
danieledamiani, line 67: we need translation but we can fix it later

danieledamiani (approved):

PR #501, opened 2021-11-09
Add APR tooltip

danieledamiani:


packages/tempus-client_v2/src/components/pool/aprTooltip/aprTooltip.tsx
```diff
@@ -0,0 +1,130 @@
...
+    if (!activePoolData) {
+      return null;
+    }
+    return NumberUtils.formatPercentage(activePoolData.fixedAPR, 2);
+  }, [activePoolData]);
+
+  const futureYieldFormatted = useMemo(() => {
+    if (activePoolData !== null) {
+      const daysToMaturity = differenceInDays(activePoolData.maturityDate, activePoolData.startDate);
+      const dailyInterest = (activePoolData.fixedAPR || 1) / 365;
```
lazarmitic, line 52: If `activePoolData.fixedAPR` is not yet fetched, we should probably wait for it to load before showing any percentage. (Or show loading spinner instead of percentage value - we should probably check with @Olafkuusik).

packages/tempus-client_v2/src/components/pool/aprTooltip/aprTooltip.tsx
```diff
@@ -0,0 +1,130 @@
...
+      return NumberUtils.formatPercentage(dailyInterest * daysToMaturity, 2);
+    }
+
+    return null;
+  }, [activePoolData]);
+
+  const lifetimeAprFormatted = useMemo(() => {
+    if (activePoolData !== null && poolRatio !== null) {
+      const daysToMaturity = differenceInDays(activePoolData.maturityDate, activePoolData.startDate);
+      const dailyInterest = (poolRatio[0] / poolRatio[1] || 1) / 365;
```
lazarmitic, line 62: Same comment as above, but for `poolRatio`.

lazarmitic (changes requested):

packages/tempus-client_v2/src/components/pool/aprTooltip/aprTooltip.tsx
danieledamiani: fixed

packages/tempus-client_v2/src/components/pool/aprTooltip/aprTooltip.tsx
danieledamiani: fixed

lazarmitic (approved):

PR #516, opened 2021-11-11
Improve APR filtering in dashboard

iamvukasin:
* Allow filling only min or max APR
* Make APR range inclusive
* Show % suffix in UI


lazarmitic:
Thanks for contributing. I will have to check with designer about '%' addition, but the code looks good!

danieledamiani (approved):
Approved if design confirms the `%` suffix

PR #527, opened 2021-11-17
Global state using hookstate library

lazarmitic:
Here is a small update presenting what we can do with [Hookstate](https://hookstate.js.org/docs/getting-started) library. I will list some of the benefits here compared to using context (our current global state management).

1. If global state updates, only components that are consuming the state are updated.
2. Minimal boilerplate code required to build a state object - it's as simple as useState react hook.
3. If state is a global state object that contains two keys A and B. And we have a component X that uses this global state by reading key A, component X will re-render only if key A is updated, if key B is updated, component X will not re-render which is a huge win if we have a large state.
4. State can be updated outside of React components as well, this is important for our use case because with this option we can update global state directly from service classes that are fetching data from contracts using rxjs or blockchain events.

There are some other minor benefits, but I think these 4 are major ones that will resolve all of the issues we currently have with context. If you have any questions please let me know. I think this is a move in right direction that will make dApp much lighter and faster.

This PR moves selectedPool from PoolData context  to new state object created by Hookstate library, I also added Maturity and Start dates for all pools in a seprate state objects that are being consumed by Term component. Major optimization here is Term component that will not re-render every time poolData changes (which is very often). For Maturity and Start dates for each pool I used map object, so when you try to get maturity date for selected pool you can just do poolsMaturityData[selectedPool], instead of going through an array with find().

Idea is to move all of the stuff from Context to Hookstate states.


packages/tempus-client_v2/src/state/PoolDataState.ts
```diff
@@ -0,0 +1,19 @@
+import { createState } from '@hookstate/core';
+import getConfig from '../utils/getConfig';
+
+// Currently selected pool (Pool Address)
```
danieledamiani, line 4: is this comment correct?

lazarmitic: Yes, this state object is used to store currently selected pool. It's not of type string, it's a Hookstate State object that contains string value (pool address) and some other utility functions for updating and reading current state.

lazarmitic:
@danieledamiani merge conflicts resolved.

danieledamiani:
In general is a good idea. I agree with points 1 to 4. I'm only not fully convinced on data structure, having a separate object for each Pool property. 

What if we do something like
```
const staticPoolData = {
  '0xfb2df41a618b44634a4e180248faaa4dc755d84a000200000000000000000036': {
      address: '0x1c5AbE736C6CCb743Bc933241AB462e6b38c6EA4',
      poolId: '0xfb2df41a618b44634a4e180248faaa4dc755d84a000200000000000000000036',
      ammAddress: '0xFb2dF41A618b44634a4E180248fAAA4dc755d84a',
      principalsAddress: '0x08C90BC3448d0928D5b6a97A6c28a4228439E7f7',
      yieldsAddress: '0x868679f952B8b89047cbC015C5c8a765F44040a8',
      yieldBearingTokenAddress: '0x1643E812aE58766192Cf7D2Cf9567dF2C37e9B7F',
      backingTokenAddress: '0x0000000000000000000000000000000000000000',
      startDate: 1633345485000,
      maturityDate: 1638190800000,
      protocol: 'lido',
      backingToken: 'ETH',
      yieldBearingToken: 'stETH',
      spotPrice: '2',
      decimalsForUI: 4,
    }
}

const dynamicPoolData = {
  '0xfb2df41a618b44634a4e180248faaa4dc755d84a000200000000000000000036': {
    userBackingTokenBalance: '0',
    userYieldBearingTokenBalance: '0',
    userPrincipalsBalance: '0',
    userYieldsBalance: '0',
    userLPTokenBalance:'0',
    spotPrice: '123',
    fixedAPR: 3,
    variableAPR: 7,
    tvl: '0',
    userBalanceUSD: '0',
    userBalanceInBackingToken: '0',
    isNegativeYield: false,
    backingTokenValueInFiat: '0',
    backingTokensAvailable: '0',
    yieldBearingTokenValueInFiat: '0',
    yieldBearingTokenValueInBackingToken: '0',
  }
}
```

lazarmitic:
Yes, I can do that. And using [Scoped State](https://hookstate.js.org/docs/scoped-state) feature from Hookstate, performance will not be affected even if we have large object states with frequent updates :D

I will amend my PR.

lazarmitic:
PR updated, there is now a single state object that will contain static pool data.

danieledamiani (approved):

PR #530, opened 2021-11-23
Move poolDataContext to Hookstate State

lazarmitic:


packages/tempus-client_v2/src/state/PoolDataState.ts
```diff
@@ -1,23 +1,110 @@
...
+  yieldBearingToken: Ticker;
+  backingTokenAddress: string;
+  yieldBearingTokenAddress: string;
+  decimalsForUI: number;
+  precision: {
+    backingToken?: number;
+  };
+}
+
+export interface StaticPoolStateData {
```
danieledamiani, line 30: a more meaningful name would be `StaticPoolDataMap` or something similar

packages/tempus-client_v2/src/state/PoolDataState.ts
```diff
@@ -1,23 +1,110 @@
...
   maturityDate: number;
   startDate: number;
+  principalsAddress: string;
+  yieldsAddress: string;
+  backingToken: Ticker;
+  yieldBearingToken: Ticker;
+  backingTokenAddress: string;
+  yieldBearingTokenAddress: string;
+  decimalsForUI: number;
+  precision: {
```
danieledamiani, line 25: should this property be called `tokenPrecision` to be consistent?

packages/tempus-client_v2/src/state/PoolDataState.ts
```diff
@@ -1,23 +1,110 @@
...
 import getConfig from '../utils/getConfig';
+import { Ticker } from '../interfaces/Token';
+import { ProtocolName } from '../interfaces/ProtocolName';
 
 // Currently selected pool (Pool Address)
 export const selectedPoolState = createState('');
 
 // Static pool data
-interface StaticPoolData {
+export interface StaticPoolData {
```
danieledamiani, line 11: actually `StaticPoolData` interface is the exact same as `TempusPool` type, so maybe no need to have both?

packages/tempus-client_v2/src/state/PoolDataState.ts
```diff
@@ -1,23 +1,110 @@
...
+
+export interface StaticPoolStateData {
+  [poolAddress: string]: StaticPoolData;
 }
 
 // Static pool data state object
-const staticPoolDataStateInitialValue: { [poolAddress: string]: StaticPoolData } = {};
+const staticPoolDataStateInitialValue: StaticPoolStateData = {};
 getConfig().tempusPools.forEach(tempusPoolConfig => {
   staticPoolDataStateInitialValue[tempusPoolConfig.address] = {
```
danieledamiani, line 37: this can be simplified as:
```
staticPoolDataStateInitialValue[tempusPoolConfig.address] = { ...tempusPoolConfig };

File: packages/tempus-client_v2/src/components/buttons/Approve.tsx [hunk 5]
```diff
@@ -55,18 +54,18 @@ const Approve: FC<ApproveButtonProps> = props => {
...
   const { userWalletAddress, userWalletSigner } = useContext(WalletContext);
   const { language } = useContext(LanguageContext);
 
   const [approveInProgress, setApproveInProgress] = useState<boolean>(false);
   const [allowance, setAllowance] = useState<BigNumber | null>(null);
 
-  const selectedPoolData = useMemo(() => {
-    return getDataForPool(selectedPool.get(), poolData);
-  }, [poolData, selectedPool]);
+  const backingToken = staticPoolData[selectedPool.get()].backingToken.attach(Downgraded).get();
```
danieledamiani, line 66: what is the rationale behind this? I think `staticPoolData` is not going to change (hence is 'static')

lazarmitic: Yes, we can omit whole staticState stuff and just use it directly from config. I realized this along the way, but it's going to take some time to change, so I just kept it. We can remove it down the line.

packages/tempus-client_v2/src/components/dashboard/dashboard.tsx
```diff
@@ -115,109 +116,102 @@ const Dashboard: FC<DashboardProps> = ({ hidden, userWalletAddress, rows, onRowA
...
+
   /**
    * If `null` is passed as `filterData`, all filters will be cleared.
    */
-  const onApplyFilter = useCallback(
-    (filterData: FilterData | null) => {
-      if (!filterData) {
-        setFilteredRows(null);
-        return;
+  const onApplyFilter = (filterData: FilterData | null) => {
```
danieledamiani, line 143: why did you remove `useCallback`?

packages/tempus-client_v2/src/components/dashboard/providers/availableToDepositFormatter.tsx
```diff
@@ -47,26 +60,28 @@ const AvailableToDepositFormatter = (props: DataTypeProvider.ValueFormatterProps
...
       );
     }
 
     return <div className="tc__dashboard__grid__avail-to-deposit__container">{content}</div>;
-  }, [poolData, parentAvailableToDeposit, showFiat]);
+  };
+  const parentAvailableToDepositFormatted = getParentAvailableToDepositFormatted();
 
-  const childAvailableToDeposit = useMemo(() => {
+  const getChildAvailableToDeposit = () => {
```
danieledamiani, line 72: why remove `useMemo`? Every time we run line `82` we rerun this function

danieledamiani (changes requested):
In general looks good but there are a few questions I'd like to be answered

packages/tempus-client_v2/src/components/dashboard/dashboard.tsx
lazarmitic: `negativeYieldPoolData` and `dynamicPoolData` reference never changes (state is created only once) and callback is not being updated properly, couple of solutions for this:

1. When setting state use spread operator to update whole state object every time - not good for performance.
2. Remove callback hook. - this component is no re-rendered a lot so even without useCallback performance is not affected a lot.
3. Use library that allows us to specify custom compare function for useCallback - this is best solution I think.

packages/tempus-client_v2/src/components/dashboard/providers/availableToDepositFormatter.tsx
lazarmitic: See response on above comment.

lazarmitic:
Added response to all comments.

packages/tempus-client_v2/src/state/PoolDataState.ts
lazarmitic: fixed

packages/tempus-client_v2/src/state/PoolDataState.ts
lazarmitic: fixed

packages/tempus-client_v2/src/state/PoolDataState.ts
lazarmitic: fixed

packages/tempus-client_v2/src/state/PoolDataState.ts
lazarmitic: fixed

danieledamiani (approved):

PR #545, opened 2021-11-29
Add slippage option

lazarmitic:


packages/tempus-client_v2/src/components/percentageInput/percentageParser.ts
```diff
@@ -0,0 +1,54 @@
+export const checkIfNumberRegExp = /\d/;
```
danieledamiani, line 1: would be nice to have some unit tests for these functions

packages/tempus-client_v2/src/components/percentageInput/percentageParser.ts
```diff
@@ -0,0 +1,54 @@
...
+    const characters = splitValue[0].split('');
+
+    // Remove non digit characters from string
+    const filteredValues = characters.filter((v: string) => checkIfNumberRegExp.test(v));
+
+    const value = Number(filteredValues.join(''));
+    if (value > 100) {
+      return '100';
+    } else {
+      return value.toString();
```
danieledamiani, line 24: if value is negative?

packages/tempus-client_v2/src/components/percentageInput/percentageInput.tsx
```diff
@@ -0,0 +1,104 @@
...
+  useEffect(() => {
+    if (defaultValue || defaultValue === '') {
+      setValue(formatValueToPercentage(defaultValue));
+    }
+  }, [defaultValue]);
+
+  const onValueChange = useCallback(
+    (event: React.ChangeEvent<HTMLInputElement>) => {
+      const currentValue = event.currentTarget.value;
+      const parsedPercentage = formatValueToPercentage(currentValue);
```
danieledamiani, line 47: I think would be better to have `formatValueToPercentage` returning an object like `{ parsedValue: 1.23, parsedValueToPercentage: '123%' }` and then use these values in `setValue` and `onChange`. Doing math operation in a component should be avoided

danieledamiani (changes requested):
a few questions and observations

packages/tempus-client_v2/src/adapters/UserBalanceDataAdapter.ts
```diff
@@ -87,10 +87,10 @@ export default class UserBalanceDataAdapter {
         }),
       )
       .pipe(
-        switchMap(([backingTokenRate, userPoolBalanceInBackingTokens]) => {
-          if (this.statisticsService && backingTokenRate && userPoolBalanceInBackingTokens) {
-            const userPoolBalanceInFiat = mul18f(userPoolBalanceInBackingTokens, backingTokenRate);
-            return of([userPoolBalanceInFiat, userPoolBalanceInBackingTokens]);
+        switchMap(([backingTokenRate, exitEstimate]) => {
+          if (this.statisticsService && backingTokenRate && exitEstimate.tokenAmount) {
```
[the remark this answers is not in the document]
lazarmitic: Yes, if call to fetch `exitEstimate` fails, `throwError()` would be called and execution will be stopped.

packages/tempus-client_v2/src/components/navbar/Settings.tsx
```diff
@@ -27,13 +37,22 @@ const Settings = () => {
...
     setSettingsMenuOpen(prevValue => !prevValue);
   }, [setSettingsMenuOpen]);
 
-  const settingsMenuAnchor = useRef<HTMLLIElement>(null);
+  const onSlippageChange = useCallback(
+    (value: string) => {
+      setUserSettings &&
+        setUserSettings(prevState => ({
+          ...prevState,
+          slippage: Number(value),
```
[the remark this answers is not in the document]
lazarmitic: Percentage input already handles value range (0-100), I will add additional check for negative values in PercentageInput parser.

packages/tempus-client_v2/src/components/percentageInput/percentageParser.ts
lazarmitic: Unit tests added.

packages/tempus-client_v2/src/components/percentageInput/percentageParser.ts
lazarmitic: Added check for negative values.

packages/tempus-client_v2/src/components/percentageInput/percentageInput.tsx
lazarmitic: Fixed

packages/tempus-client_v2/src/components/percentageInput/percentageParser.test.ts
```diff
@@ -0,0 +1,22 @@
+import { formatValueToPercentage } from './percentageParser';
+
+describe('formatValueToPercentage()', () => {
+  [
+    { value: '', expected: { parsedValue: '', percentageValue: '' } },
```
danieledamiani, line 5: is there a case when `parsedValue` is different from `percentageValue`?

danieledamiani (changes requested):
one more question

packages/tempus-client_v2/src/components/percentageInput/percentageParser.test.ts
lazarmitic: oh yeah, I guess you are right, we don't have large numbers in PercentageParser like `100,100.20` (in this case `parsedValue = 100,100.20` and `percentageValue = 100100.20`). I will remove this code and remove regex from `onChange()` callback.

lazarmitic: Fixed.

packages/tempus-client_v2/src/components/percentageInput/percentageInput.tsx
```diff
@@ -0,0 +1,104 @@
...
+      if (parsedPercentage || parsedPercentage === '') {
+        setValue(parsedPercentage);
+      }
+      onChange && onChange(parsedPercentage);
```
danieledamiani, line 51: if `parsedPercentage` is falsy, is it ok to pass it to `onChange`? Otherwise we could move line 51 just below line 49

danieledamiani (changes requested):
One more!

lazarmitic:
Fixed :D

danieledamiani (approved):

PR #546, opened 2021-11-29
Add Wallet Selector - MetaMask and WalletConnect

danieledamiani:


packages/tempus-client_v2/src/components/wallet/UserWallet.ts
```diff
@@ -0,0 +1,3 @@
+type UserWallet = 'MetaMask' | 'WalletConnect';
```
lazarmitic, line 1: This type definition file should probably go to interfaces directory.

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -0,0 +1,306 @@
...
+            }));
+        });
+      }
+    };
+    if (!active) {
+      checkConnection();
+    }
+  }, [active, setWalletData, activate, requestNetworkChange]);
+
+  useEffect(() => {
```
lazarmitic, line 225: What happens if user is connected using WalletConnect and MetaMask is not installed, this useEffect will set `userWalletConnected` to `false`, right?

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -0,0 +1,306 @@
...
+              <Spacer size={10} />
+            </>
+          )}
+
+          {!connecting && selectedWallet && active && (
+            <div className="tc__connect-wallet-button" onClick={onSelectWallet}>
+              <AccountBalanceWalletIcon />
+              <Spacer size={4} />
+              {pendingTransactions.length === 0 && (
+                <Typography variant="h5">{active ? shortenedAccount : getText('connectWallet', language)}</Typography>
```
lazarmitic, line 291: `active` flag will always be true here, no need to check for it, always show `shortenedAccount`.

lazarmitic (changes requested):

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
danieledamiani: right, fixed

packages/tempus-client_v2/src/components/wallet/UserWallet.ts
danieledamiani: moved

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
danieledamiani: actually this bit is not needed, removed

lazarmitic (approved):

PR #547, opened 2021-11-30
Hide basic section in sidebar if pool does not have any liquidity

lazarmitic:


packages/tempus-client_v2/src/components/operations/Operations.tsx
```diff
@@ -31,18 +31,31 @@ const Operations = () => {
...
+  const userPrincipalsBalance = dynamicPoolData[selectedPool.get()].userPrincipalsBalance.get();
+  const userYieldsBalance = dynamicPoolData[selectedPool.get()].userYieldsBalance.get();
+  const userLPBalance = dynamicPoolData[selectedPool.get()].userLPTokenBalance.get();
+
   const handleWithdraw = useCallback(() => {
     setSelectedView('deposit');
   }, []);
 
+  /**
+   * If pool does not have any liquidity, set Mint view as a default one
```
danieledamiani, line 44: then doesn't make sense to have 'mint' as starting value at line 32?

danieledamiani (changes requested):
a comment

packages/tempus-client_v2/src/components/operations/Operations.tsx
lazarmitic: Reason for using 'Deposit' as a default one (at the start of the function) is because most of the time (99%) pool will have liquidity and this useEffect will be skipped.

danieledamiani (approved):

PR #554, opened 2021-12-02
Partial withdrawal

lazarmitic:


packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
```diff
@@ -80,38 +128,85 @@ const Withdraw: FC<WithdrawOutProps> = ({ onWithdraw }) => {
...
+        ),
+      );
+      const minYieldsStaked = estimatedWithdrawData.yieldsStaked.sub(
+        mul18f(
+          estimatedWithdrawData.yieldsStaked,
+          ethers.utils.parseUnits((slippage / 100).toString(), getTokenPrecision(selectedPoolAddress, 'yields')),
+        ),
+      );
+
+      let minRate = BigNumber.from('0');
```
danieledamiani, line 187: if `principalsAmountParsed` is equal to `yieldsAmountParsed` then `minRate` is `0`, is that correct?

lazarmitic: Yes, in that case there is no swap and rate is not needed. I will verify this with Djordje.

packages/tempus-client_v2/src/services/StatisticsService.ts
```diff
@@ -258,14 +257,7 @@ class StatisticsService {
...
 
-    let maxLeftoverShares: BigNumber;
-    try {
-      maxLeftoverShares = await this.tempusAMMService.getMaxLeftoverShares(tempusAmmAddress, principalAmount, lpAmount);
-    } catch (error) {
-      console.error('StatisticsService - estimateExitAndRedeem() - Failed to fetch max leftover shares!');
-      return Promise.reject();
-    }
-
+    let maxLeftoverShares = this.tempusAMMService.getMaxLeftoverShares(principalAmount, yieldsAmount, lpAmount);
```
danieledamiani, line 260: no `try...catch` block anymore?

danieledamiani (changes requested):
looks ok, 2 questions pending

packages/tempus-client_v2/src/services/StatisticsService.ts
lazarmitic: It's not an async function any more, and `principalAmount, yieldsAmount, lpAmount` are always defined, so it's much less likely it's going to fail.

packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
lazarmitic: I missed one thing here, I need to include LP tokens in principal and yield amounts, because when we withdraw, contract converts LP tokens to principals and yields, and then does the swap.

lazarmitic: fixed

packages/tempus-client_v2/src/services/StatisticsService.ts
danieledamiani: `tempusAMMService.getMaxLeftoverShares` has a condition `if (!this.eRC20TokenServiceGetter) {` which may cause the function to throw an error, in that case it'll break the app.

packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
lazarmitic: Yes, I spoke with Djordje, and based on what he told me I fixed the code.

packages/tempus-client_v2/src/services/StatisticsService.ts
lazarmitic: Oh, I don't think that check is needed any more, I will remove it.

lazarmitic: Fixed

danieledamiani (approved):

PR #557, opened 2021-12-06
Add pool route

lazarmitic:


packages/tempus-client_v2/src/components/routes/PoolRoute.tsx
```diff
@@ -0,0 +1,26 @@
...
+import Operations from '../operations/Operations';
+
+const PoolRoute = () => {
+  const selectedPool = useHookstate(selectedPoolState);
+
+  const params = useParams();
+
+  useEffect(() => {
+    if (params.poolAddress) {
+      // TODO - Refactor whole app to use pool address from URL instead of using it from state
```
danieledamiani, line 14: is it necessary? I think we should get pool address once from params and then we're good

danieledamiani (approved):

danieledamiani (approved):

PR #562, opened 2021-12-06
Add Wallet overview and transaction history

danieledamiani:


packages/tempus-client_v2/src/components/wallet/Wallet.scss
```diff
@@ -30,4 +30,14 @@
...
   padding-right: 15px;
   box-shadow: -2px 0px 2px rgba(0, 0, 0, 0.25);
+  cursor: pointer;
+
+  &.tc__connect-wallet-button__connected:hover {
+    background-color: $link-color;
+    color: $inverted-text-color;
+
+    > div {
+      color: $inverted-text-color !important;
```
lazarmitic, line 40: You can also use Typography inverted prop for this.

lazarmitic (approved):

packages/tempus-client_v2/src/components/wallet/Wallet.scss
danieledamiani: how to do it on mouse over?

lazarmitic: Ah, yeah, you are right. Inverted does not take into account mouse over state.

PR #571, opened 2021-12-07
Change "Liquidation Value" to "Current Value"

lazarmitic:


packages/tempus-client_v2/src/components/profitLoss/ProfitLoss.tsx
```diff
@@ -71,7 +71,7 @@ const ProfitLoss = () => {
       <div className="tc__profitLoss__body">
         <div className="tf__flex-row-space-between tc__underline">
           <Typography variant="card-body-text" color="title">
-            {getText('liquidationValue', language)}
+            {getText('currentValue', language)}
```
danieledamiani, line 74: we could keep the same variable and change only translation...

packages/tempus-client_v2/src/localisation/it.ts
```diff
@@ -78,7 +78,7 @@ const it: { [word in Words]: string } = {
   execute: 'Esegui',
   executing: 'In Esecuzione',
   profitLoss: 'Profitti & Perdite',
-  liquidationValue: 'Valore di Liquidazione',
+  currentValue: 'Current Value',
```
danieledamiani, line 81: that would be `Valore corrente`

danieledamiani (approved):

packages/tempus-client_v2/src/components/profitLoss/ProfitLoss.tsx
lazarmitic: I changed it because "liquidationValue": "Current Value" seemed a bit wierd.

packages/tempus-client_v2/src/localisation/it.ts
lazarmitic: fixed

PR #573, opened 2021-12-07
Disable CircleCI build step

lazarmitic:


lazarmitic:
I guess we can also mark 'Build' step as not required, and keep code intact.

PR #589, opened 2021-12-08
Fix Withdraw execute disabled state logic

lazarmitic:


packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
```diff
@@ -354,17 +354,33 @@ const Withdraw: FC<WithdrawOutProps> = ({ onWithdraw }) => {
     const yieldsAmountZero = !yieldsAmount || isZeroString(yieldsAmount);
     const lpTokenAmountZero = !lpTokenAmount || isZeroString(lpTokenAmount);
 
+    const principalsExceedsBalance = ethers.utils
+      .parseEther(principalsAmount || '0')
```
danieledamiani, line 358: we should use `parseUnits`

danieledamiani (changes requested):
a comment

packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
lazarmitic: ah, yes, I will do that.

lazarmitic: Fixed.

danieledamiani (approved):

PR #591, opened 2021-12-08
Update slippage selector

danieledamiani:


packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
```diff
@@ -175,16 +175,18 @@ const Withdraw: FC<WithdrawOutProps> = ({ onWithdraw }) => {
         getTokenPrecision(selectedPoolAddress, 'lpTokens'),
       );
 
+      const actualSlippage = (autoSlippage ? 100 : slippage / 100).toString();
```
lazarmitic, line 178: if auto is set to true, slippage needs to be set to 1 (100%)

lazarmitic (changes requested):

packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
danieledamiani: right!!

lazarmitic (approved):

PR #603, opened 2021-12-10
Withdraw function interface change

lazarmitic:


packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
```diff
@@ -218,7 +198,8 @@ const Withdraw: FC<WithdrawOutProps> = ({ onWithdraw }) => {
         lpTokenAmountParsed,
         minPrincipalsStaked,
         minYieldsStaked,
-        minRate,
+        estimatedWithdrawData.yieldsRate,
+        ethers.utils.parseEther(actualSlippage),
```
danieledamiani, line 202: please use `parseUnit`

danieledamiani (changes requested):
a comment

packages/tempus-client_v2/src/components/withdraw/Withdraw.tsx
lazarmitic: It's hardcoded on contract side to use 18 decimals for slippage, slippage is not a token, so no need to use token precision here.

danieledamiani: then for consistency let's do `parseUnit(actualSlippage, 18)`

lazarmitic: fixed

danieledamiani (approved):

PR #608, opened 2021-12-13
Update config for mainnet

danieledamiani:


packages/tempus-client_v2/src/config/config.ts
```diff
@@ -1,19 +1,59 @@
...
+// const GOERLI_ALCHEMY_KEY = process.env.REACT_APP_GOERLI_ALCHEMY_KEY || '';
+// const GOERLI_ALCHEMY_CONFIG = {
+//   networkUrl: 'https://eth-goerli.alchemyapi.io/v2/RUE8LXWaELu4_nyMgfK99eX3DWoxLtDw',
+//   networkName: 'goerli',
+//   alchemyKey: GOERLI_ALCHEMY_KEY,
+// } as Config;
+
+const MAINNET_ALCHEMY_KEY = process.env.REACT_APP_MAINNET_ALCHEMY_KEY || '';
+const MAINNET_ALCHEMY_CONFIG = {
+  networkUrl: 'https://eth-mainnet.alchemyapi.io/v2/CikBd5PRY8HTC1rBQw1dBq5HlOfXwe0O',
```
lazarmitic, line 12: we should probably hide alchemy key from this URL as well.

packages/tempus-client_v2/src/services/NotificationService.ts
```diff
@@ -117,10 +116,11 @@ class NotificationService {
...
 
 export const generateEtherscanLink = (tx: string) => {
-  const config = getConfig();
-
+  // const config = getConfig();
   // TODO - No need to specify network name for mainnet
-  return `https://${config.networkName}.etherscan.io/tx/${tx}`;
+  // return `https://${config.networkName}.etherscan.io/tx/${tx}`;
+
+  return `https://.etherscan.io/tx/${tx}`;
```
lazarmitic, line 123: there is an extra . before etherscan.

lazarmitic (changes requested):

packages/tempus-client_v2/src/config/config.ts
danieledamiani: right

packages/tempus-client_v2/src/services/NotificationService.ts
danieledamiani: good spot

danieledamiani:
updated

lazarmitic (approved):

PR #626, opened 2021-12-14
Update LP APR calculation to use 7 days events

danieledamiani:


lazarmitic:
What we need to change here as well is the scale factor, when scaling to one year instead of using 365, we need to get time of oldest event (or block) that we queried, do `365 / (currentTime - pastBlockTime)`, and use result of that when scaling to one year.

danieledamiani:
> What we need to change here as well is the scale factor, when scaling to one year instead of using 365, we need to get time of oldest event (or block) that we queried, do 365 / (currentTime - pastBlockTime), and use result of that when scaling to one year.

Refactored

packages/tempus-client_v2/src/services/VariableRateService.ts
```diff
@@ -154,7 +156,16 @@ class VariableRateService {
...
+    const { startDate } = poolConfig;
+
+    const earlierBlock = await provider.getBlock(
+      latestBlock.number - Math.floor(SECONDS_IN_A_WEEK / BLOCK_DURATION_SECONDS),
+    );
+
+    const laterBlock = startDate >= earlierBlock.timestamp ? startDate : earlierBlock.timestamp;
+    const hoursBetweenLatestAndLater = Math.floor((latestBlock.timestamp * 1000 - laterBlock) / (60 * 60 * 1000));
+
+    const fetchEventsFromBlock = latestBlock.number - laterBlock;
```
lazarmitic, line 168: You are mixing timestamp and block number here. You can fetch events from all blocks, even if pool was not deployed at the time - only reason to calculate if poolStartTime is after oldest block is for scale factor calculation.

packages/tempus-client_v2/src/services/VariableRateService.ts
```diff
@@ -154,7 +156,16 @@ class VariableRateService {
...
 
-    const fetchEventsFromBlock = latestBlock.number - Math.floor(SECONDS_IN_A_DAY / BLOCK_DURATION_SECONDS);
+    const { startDate } = poolConfig;
+
+    const earlierBlock = await provider.getBlock(
+      latestBlock.number - Math.floor(SECONDS_IN_A_WEEK / BLOCK_DURATION_SECONDS),
+    );
+
+    const laterBlock = startDate >= earlierBlock.timestamp ? startDate : earlierBlock.timestamp;
+    const hoursBetweenLatestAndLater = Math.floor((latestBlock.timestamp * 1000 - laterBlock) / (60 * 60 * 1000));
```
lazarmitic, line 166: latestBlock.timestamp multiplied by 1000 to convert it to `ms`, and `laterBlock` remains to be in seconds, in case `startDate >= earlierBlock.timestamp` is false.

packages/tempus-client_v2/src/services/VariableRateService.ts
```diff
@@ -224,7 +235,11 @@ class VariableRateService {
     });
 
     // Scale accumulated fees to 1 year duration
-    const scaledFees = mul18f(totalFees, ethers.utils.parseEther(DAYS_IN_A_YEAR.toString()));
+
+    const scaledFees = mul18f(
+      div18f(totalFees, ethers.utils.parseEther(hoursBetweenLatestAndLater.toString())),
```
lazarmitic, line 240: I'm not sure why are you dividing this, you can just change `ethers.utils.parseEther(DAYS_IN_A_YEAR.toString()` to proper scale factor. Instead of multiplying by 365 like before (you get number of hours `hoursBetweenLatestAndLater` and just divider number of hours in a year with that).

lazarmitic (changes requested):

packages/tempus-client_v2/src/services/VariableRateService.ts
danieledamiani: fixed

packages/tempus-client_v2/src/services/VariableRateService.ts
danieledamiani: refactored...

packages/tempus-client_v2/src/services/VariableRateService.ts
danieledamiani: `scaledFees = (totalFees / hoursBetweenLatestAndLater  ) * HOURS_IN_A_YEAR` should be the fees scaled to a year?

lazarmitic: yeah, you can do it like that as well.. (the scaling)

danieledamiani:
I had to revert to
`const earlierBlock = await provider.getBlock(
      latestBlock.number - Math.floor(SECONDS_IN_A_WEEK / BLOCK_DURATION_SECONDS),
    );`

lazarmitic (approved):

PR #627, opened 2021-12-14
Properly fetch user balance for pools when wallet is connected

lazarmitic:


packages/tempus-client_v2/src/providers/getBalanceProvider.ts
```diff
@@ -0,0 +1,17 @@
+import UserBalanceProvider, { UserBalanceProviderParams } from './balanceProvider';
+
+// Key - User wallet address
+// Value - Provider instance
+const userBalanceProviderMap = new Map<string, UserBalanceProvider>();
+const getUserBalanceProvider = (params: UserBalanceProviderParams): UserBalanceProvider => {
+  let userBalanceProvider = userBalanceProviderMap.get(params.userWalletAddress);
```
danieledamiani, line 7: in `UserBalanceProvider` we call `this.destroy()` inside `init()`, so what's the reason to save a `UserBalanceProvider` instance?

packages/tempus-client_v2/src/providers/balanceProvider.tsx
```diff
@@ -1,134 +1,106 @@
...
+      lpTokenService.balanceOf(this.userWalletAddress),
+      statisticsService.getRate(poolConfig.backingToken),
+    ]);
+
+    const exitEstimate = await statisticsService.estimateExitAndRedeem(
+      poolConfig.ammAddress,
+      lpTokenBalance,
+      principalsBalance,
+      yieldsBalance,
+      true,
```
danieledamiani, line 83: what's `true`?

packages/tempus-client_v2/src/providers/balanceProvider.tsx
```diff
@@ -1,134 +1,106 @@
...
-          }),
-        );
-
-        subscriptions$.add(
-          stream$.subscribe((balances: BigNumber[]) => {
-            updateUserBalanceForPool(poolConfig.address, balances);
-          }),
-        );
-      }
+      this.tokenContracts.push(tpsContract, tysContract);
```
danieledamiani, line 41: you're not pushing `lpTokenContract`, is that intentional?

danieledamiani (changes requested):
a few questions

packages/tempus-client_v2/src/providers/getBalanceProvider.ts
lazarmitic: Are you suggesting to change it to static?

packages/tempus-client_v2/src/providers/balanceProvider.tsx
lazarmitic: I will add a variable.

packages/tempus-client_v2/src/providers/balanceProvider.tsx
lazarmitic: ah, good catch, I will fix it.

lazarmitic: fixed

packages/tempus-client_v2/src/providers/balanceProvider.tsx
lazarmitic: fixed

packages/tempus-client_v2/src/providers/getBalanceProvider.ts
danieledamiani: I'm saying that maybe we don't need to save the instance in `userBalanceProviderMap` at all

lazarmitic: oh, yeah, maybe. Hmm, yeah, I think we can do that. I will try it

lazarmitic: fixed

packages/tempus-client_v2/src/providers/getBalanceProvider.ts
```diff
@@ -2,16 +2,8 @@ import UserBalanceProvider, { UserBalanceProviderParams } from './balanceProvide
 
 // Key - User wallet address
```
danieledamiani, line 3: and these comments?

danieledamiani (dismissed):

packages/tempus-client_v2/src/providers/getBalanceProvider.ts
lazarmitic: ahh, missed those, I will remove them

danieledamiani (approved):

PR #630, opened 2021-12-15
Properly set min TYS rate for fixed deposit

lazarmitic:


packages/tempus-client_v2/src/adapters/PoolDataAdapter.ts
```diff
@@ -407,6 +408,14 @@ export default class PoolDataAdapter {
 
     try {
       if (yieldType === 'Fixed') {
+        const tysRate = await this.getExpectedReturnForShareToken(
+          tempusAMM,
+          ethers.utils.parseUnits('1', yieldsPrecision),
```
danieledamiani, line 413: what's `1` for?

danieledamiani (approved):

packages/tempus-client_v2/src/adapters/PoolDataAdapter.ts
lazarmitic: fixed

danieledamiani (approved):

PR #631, opened 2021-12-15
Update wrong network message to "MainNet"

danieledamiani:


packages/tempus-client_v2/src/localisation/en.ts
```diff
@@ -14,9 +14,9 @@ const en: { [word in Words]: string } = {
     'By connecting your wallet, you agree to be bound by the <a href="https://tempus.finance/terms-of-service" target="_blank">Tempus Terms of Service</a> and acknowledge that you have read and understand the <a href="https://tempus.finance/disclaimer" target="_blank">Tempus Protocol Disclaimer</a>.',
   metamaskConnected: 'MetaMask connected',
   changeNetworkRejected: 'Request to change network rejected by user',
-  changeNetworkRejectedExplain: 'In order to use the app, please connect using Goerli network',
+  changeNetworkRejectedExplain: 'In order to use the app, please connect using MainNet network',
```
lazarmitic, line 17: I think the casing should be 'Mainnet', Metamask uses that.

lazarmitic (changes requested):

packages/tempus-client_v2/src/localisation/en.ts
danieledamiani: fixed

lazarmitic (approved):

PR #632, opened 2021-12-15
Use same fixed APR fetch function for dashboard and deposit and fix UI

lazarmitic:


packages/tempus-client_v2/src/providers/fixedAPRProvider.tsx
```diff
@@ -42,11 +45,17 @@ const FixedAPRProvider = () => {
         }
 
         try {
-          // Get fees for Tempus Pool
-          const fixedAPR = await tempusAMMSService.getFixedAPR(tempusPool.ammAddress, tempusPool.principalsAddress);
+          const spotPrice = ethers.utils.parseUnits('1', getTokenPrecision(tempusPool.address, 'backingToken'));
+
+          const fixedAPR = await poolDataAdapter.getEstimatedFixedApr(
+            spotPrice,
+            true,
```
danieledamiani, line 52: what's `true`?

danieledamiani (approved):

packages/tempus-client_v2/src/providers/fixedAPRProvider.tsx
lazarmitic: I will create a variable.

lazarmitic: fixed

danieledamiani (approved):

PR #654, opened 2021-12-23
Update deposit yield cards style

lazarmitic:


packages/tempus-client_v2/src/components/icons/SelectedIcon.tsx
```diff
@@ -0,0 +1,10 @@
+const SelectedIcon = () => {
+  return (
```
danieledamiani, line 2: since we're using a react component to render an svg, could we use only one and pass a props `selected` which changes the color of the inner circle from white/transparent to blue?

danieledamiani (dismissed):
A minor comment

packages/tempus-client_v2/src/components/icons/SelectedIcon.tsx
lazarmitic: Yes, I will do that!

lazarmitic: fixed

danieledamiani (approved):

PR #660, opened 2021-12-28
Simplify and fix fixedRateAPR calculation.

mijovic:


danieledamiani (dismissed):

packages/tempus-client_v2/src/adapters/PoolDataAdapter.ts
```diff
@@ -805,20 +805,13 @@ export default class PoolDataAdapter {
...
         tempusPoolAddress,
         tokenAmount,
-        interestRate,
+        isBackingToken
       );
 
-      const ratio = div18f(principals, backingAmount);
+      const ratio = div18f(principals, estimatedMintedShares);
       const pureInterest = ratio.sub(BigNumber.from(ONE_ETH_IN_WEI));
       return mul18f(pureInterest, scaleFactor);
```
lazarmitic, line 816: I think that another problem here is that scale factor takes into account whole pool duration instead of remaining pool duration. When I change scale factor to use remaining pool duration everything matches properly (estimated principals received and fixed APR percentage.)

mijovic: Hm, I think you are missing one thing here. Value of one Yield token is calculated based on whole duration of the pool, so scale factor is correct.
APR for the whole lenght is numPrincipalsFromSwap / principalsMinted, which is exactly how I calculate it here... Then it just scales up. 
If you deploy this locally, you will clearly see that number are correct

lazarmitic (approved):

PR #665, opened 2022-01-01
Update profit loss chart when user balance for pool changes

lazarmitic:


packages/tempus-client_v2/src/components/profitLoss/profitLossChart/ProfitLossChart.tsx
```diff
@@ -41,6 +44,39 @@ const ProfitLossChart = () => {
...
+   */
+  useEffect(() => {
+    if (!userBalanceUSD || chartData.length === 0) {
+      return;
+    }
+
+    const currentChartData = [...chartData];
+
+    const lastChartEntry = currentChartData.pop();
+    if (!lastChartEntry) {
```
danieledamiani, line 58: this value can be `undefined` only if `currentChartData.length === 0` but this is always false as you're checking this condition at line `51`. Is there a case where the last value of `chartData` returns a `null` value?

danieledamiani (changes requested):
a comment

packages/tempus-client_v2/src/components/profitLoss/profitLossChart/ProfitLossChart.tsx
lazarmitic: Yeah, but TypeScript does not get that, return type on pop is <ChartDataPoint | undefined>. So I have to check, otherwise TypeScript complains.

danieledamiani (approved):

PR #676, opened 2022-01-11
Use pool dates from config - Closes #671

### LINKED ISSUE #671: Use pool start and maturity data info from static store instead of fetching it from contract in `getEstimatedFixedApr`

danieledamiani:


ShawTim:
looks good but the unit test fails for `getEstimatedFixedApr()`. it's testing with mock address but `staticPoolDataState` contains real data so it fails to get the tempus pool data

danieledamiani:
> looks good but the unit test fails for `getEstimatedFixedApr()`. it's testing with mock address but `staticPoolDataState` contains real data so it fails to get the tempus pool data

yes I'll find another way to test it

ShawTim (approved):

PR #685, opened 2022-01-12
unit test for src/services/VariableRateService.ts

ShawTim:
refactor calculateFees() to break down into smaller functions


ShawTim:
this is outdated. replace this by #761 with some  other unit tests

PR #684, opened 2022-01-12
Integrate fixed apr rari

danieledamiani:

PR #687, opened 2022-01-14
Update token icons

danieledamiani:
~~merge only after #684~~


packages/tempus-client_v2/src/components/tokenIcon/index.tsx
```diff
@@ -1,27 +1,49 @@
...
+  const iconHeight = useMemo(() => {
+    if (height) {
+      return height;
+    }
+
+    return large ? 40 : 20;
+  }, [height, large]);
+
+  const viewBox = useMemo(() => {
+    return large ? '0 0 40 40' : '0 0 20 20';
```
lazarmitic, line 35: If user passes custom width and height (for example 17x17) viewbox will not be correct.

lazarmitic (changes requested):

packages/tempus-client_v2/src/components/tokenIcon/index.tsx
danieledamiani: Fixed

lazarmitic (approved):

PR #689, opened 2022-01-14
Fix token rate precision issue

lazarmitic:


packages/tempus-client_v2/src/constants.ts
```diff
@@ -37,6 +37,12 @@ export const dashboardChildMaturityFormat = 'd MMMM yyyy';
 
 export const POLLING_INTERVAL = 30 * 1000;
 
+export const tokenPrecision: { [ticker: string]: number } = {
```
danieledamiani, line 40: can't we retrieve this value from config?

lazarmitic: we can and should, but it will take some time to refactor getRate to accept one more parameter for token precision - I added a TODO comment in statistics service for now, I will try to refactor this on monday (or today if I get some leftover time).

danieledamiani (approved):

ShawTim (approved):
works great

PR #803, opened 2022-01-21
Add Spanish translation - Closes #774

### LINKED ISSUE #774: Add Spanish translation

danieledamiani:


packages/tempus-client_v2/src/localisation/es.ts
```diff
@@ -0,0 +1,160 @@
+import Words from './words';
+
+const en: { [word in Words]: string } = {
```
ShawTim, line 3: we may want to use `es` instead of `en` here, it works though.

ShawTim (approved):

ShawTim (commented):
build fails due to missing key for `totalValueLocked`
reference:
https://github.com/tempus-finance/tempus-app/commit/82e1000400adffd8542a5392335319ff74b67a65

packages/tempus-client_v2/src/localisation/es.ts
danieledamiani: good catch

danieledamiani:
PR fixed and updated

lazarmitic (approved):

PR #827, opened 2022-01-25
Fix Fantom Chainlink ENS

danieledamiani:


packages/tempus-client_v2/src/adapters/PoolDataAdapter.ts
```diff
@@ -146,7 +147,7 @@ export default class PoolDataAdapter {
               from(yieldBearingTokenService.balanceOf(userWalletAddress)),
               from(principalsTokenService.balanceOf(userWalletAddress)),
               from(yieldsTokenService.balanceOf(userWalletAddress)),
-              from(this.statisticService.getRate(backingTokenTicker)),
+              from(this.statisticService.getRate(selectedChainState.get(), backingTokenTicker)),
```
lazarmitic, line 150: We should pass in selected chain as a param here. If some component is using `retrieveBalances()` function inside useEffect, by passing chain as a param from that component, we force developer to put selected chain inside useEffect dependencies, that will cause balance re-fetch every time chain is changed.

Logic I use is, never read state outside of react component. If we follow this rule, all relevant components will re-render when piece of state that component uses is updated.

lazarmitic (changes requested):

packages/tempus-client_v2/src/adapters/PoolDataAdapter.ts
danieledamiani: right, fixed

lazarmitic (approved):

PR #829, opened 2022-01-26
Multiple network support for config

lazarmitic:


packages/tempus-client_v2/src/state/NetworkState.ts
```diff
@@ -0,0 +1,6 @@
+import { createState } from '@hookstate/core';
+
+export type Networks = 'ethereum-mainnet' | 'fantom-mainnet';
```
danieledamiani, line 3: I'd prefer to keep `Chain` to differentiate from `Network` (`mainnet`, `goerli` and so on). Also values should be `ethereum`, not `ethereum-mainnet`

packages/tempus-client_v2/src/adapters/getPoolDataAdapter.ts
```diff
@@ -6,29 +6,38 @@ import getTempusAMMService from '../services/getTempusAMMService';
...
+      tempusAMMService: getTempusAMMService(chain, getDefaultProvider(chain)),
+      vaultService: getVaultService(chain, getDefaultProvider(chain)),
+      chain,
       eRC20TokenServiceGetter: getERC20TokenService,
     });
+    poolDataAdapters.set(chain, poolDataAdapter);
+  }
+
+  const poolDataAdapter = poolDataAdapters.get(chain);
+  if (!poolDataAdapter) {
```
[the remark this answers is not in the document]
lazarmitic: Yes, but TypeScript complains because .get() function from map returns `PoolDataAdapter | undefined`, so I have to check to avoid TS errors.

packages/tempus-client_v2/src/utils/getConfig.ts
```diff
@@ -19,8 +19,38 @@ export default function getConfig(): Config {
   }
 }
 
+export function getChainConfig(chain: Chain): ChainConfig {
```
danieledamiani, line 22: this function should leverage `getConfig()` and avoid code repetition. Something like 
```
const localConfig = getConfig();
const networkConfig = localConfig[chain];
```

packages/tempus-client_v2/src/utils/getConfig.ts
```diff
@@ -19,8 +19,38 @@ export default function getConfig(): Config {
...
+    if (!networkConfig) {
+      throw new Error(`Failed to get config for ${chain} chain from cookie!`);
+    }
+    return networkConfig;
+  }
+}
+
 export function getConfigForPoolWithId(poolId: string): TempusPool {
-  const poolConfig = getConfig()[selectedChainState.get()].tempusPools.find(tempusPool => {
+  const config = getConfig();
```
danieledamiani, line 46: this works but `config` shadows the one imported at line 3, would be better to rename it to avoid any confusion

packages/tempus-client_v2/src/services/NotificationService.ts
```diff
@@ -117,9 +117,10 @@ class NotificationService {
   }
 }
 
-export const generateEtherscanLink = (tx: string) => {
-  const config = getConfig()[selectedChainState.get()];
+export const generateEtherscanLink = (tx: string, chainName: Chain) => {
+  const config = getChainConfig(chainName);
 
+  // TODO - Handle Fantom chain
```
danieledamiani, line 123: maybe this one `https://ftmscan.com/`

danieledamiani (approved):
just a few minor comments, lgtm

packages/tempus-client_v2/src/services/NotificationService.ts
lazarmitic: Yes, that's the etherscan of fantom, we can fix this as a separate PR, this one is huge as it is :D

lazarmitic:
fixed all requested changes.

danieledamiani (approved):

PR #831, opened 2022-01-26
show tooltip for LP APR breakdown, tooltip refactor

ShawTim:
image

- refactor `components/infoTooltip` to support customized anchor and React node as tooltip content (this should be an atomic components that will be reused by `feesTooltip` and `aprTooltip` as well, but it's out of scope of this task)
- show arrow for tooltip (we are using [material UI v4](https://v4.mui.com/components/popper/) which is using [Popper v1](https://popper.js.org/docs/v1/#modifiers..arrow). to make arrow working we need to implement the arrow CSS which can be referred [here](https://github.com/mui-org/material-ui/blob/4f2a07e140c954b478a6670c009c23a59ec3e2d4/docs/src/pages/components/popper/ScrollPlayground.js)
- `PoolDataState` to support `tempusFees`
- added translation

### DIFF
--- a/packages/tempus-client_v2/src/components/dashboard/providers/variableAPRFormatter.tsx
+++ b/packages/tempus-client_v2/src/components/dashboard/providers/variableAPRFormatter.tsx
@@ -1,3 +1,4 @@
+import React, { useMemo } from 'react';
 import { CircularProgress } from '@material-ui/core';
 import { Downgraded, useState as useHookState } from '@hookstate/core';
 import { ZERO } from '../../../constants';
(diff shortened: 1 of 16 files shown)


packages/tempus-client_v2/src/components/dashboard/providers/variableAPRFormatter.tsx
```diff
@@ -11,21 +12,21 @@ import {
...
-  const getAPR = () => {
+  const apr = useMemo(() => {
     if (isChild) {
       return getChildAPR(row.id, dynamicPoolData);
     } else {
       return getParentAPR(row.id, staticPoolData, dynamicPoolData);
     }
-  };
-  const apr = getAPR();
+  }, [isChild, row.id, staticPoolData, dynamicPoolData]);
```
lazarmitic, line 29: If anything inside dynamicPoolData state changes, this use memo will not re-run, and APR will stay the same, that's why  we decided to skip using useMemo and run getAPR every time this re-renders (which is not a lot).

ShawTim: thanks, skipped `useMemo` in latest commit

danieledamiani (approved):

PR #838, opened 2022-01-28
use and lock 6.0.9 for react-error-overlay

ShawTim:


lazarmitic:
There are some build warnings. Needs to be fixed.

ShawTim:
> There are some build warnings. Needs to be fixed.

good catch, fixed.

lazarmitic (approved):

PR #857, opened 2022-01-31
Use per chain average block times

lazarmitic:


ShawTim (approved):

packages/tempus-client_v2/src/interfaces/Config.ts
```diff
@@ -10,6 +10,7 @@ export interface ChainConfig {
   networkName: string;
   alchemyKey: string;
   chainId: number;
+  averageBlockTime: number;
```
danieledamiani, line 13: `averageBlockTime` in what time unit and against what time frame? I'd rename the variable to be a bit more descriptive

packages/tempus-client_v2/src/config/config.ts
```diff
@@ -89,6 +89,7 @@ const config: Config = {
     networkName: 'homestead',
     alchemyKey: ETHEREUM_MAINNET_ALCHEMY_KEY,
     chainId: 1,
+    averageBlockTime: 13.2,
```
danieledamiani, line 92: if this is expressed in seconds, should we convert it to milliseconds so to avoid javascript float number calculation?

lazarmitic: Most of the time we use this block time to calculate how many block we have in a single day (day_seconds / block_seconds), since we are doing division here, I don't think we will get much better precision by using milliseconds.

packages/tempus-client_v2/src/interfaces/Config.ts
lazarmitic: Added JSDoc comment to better describe variable and unit.

danieledamiani (approved):

PR #866, opened 2022-02-01
Fix missing percentage sign in current position in case user balance is missing

lazarmitic:


packages/tempus-client_v2/src/components/currentPosition/CurrentPosition.tsx
```diff
@@ -121,11 +121,11 @@ const CurrentPosition: FC<CurrentPositionInProps> = ({ language }) => {
 
   const unstakedPrincipalsPercentage = useMemo(() => {
     if (!userPrincipalsBalance || !lpTokenPrincipalReturnBalance) {
-      return 0;
+      return '0%';
```
danieledamiani, line 124: we should use `NumberUtils.formatPercentage`

danieledamiani (commented):
a comment

packages/tempus-client_v2/src/components/currentPosition/CurrentPosition.tsx
lazarmitic: Makes sense, I will do that now!

lazarmitic: Fixed

danieledamiani (approved):

PR #892, opened 2022-02-03
Fix UI issue relative to Spanish translation - closes  #899

danieledamiani:


packages/tempus-client_v2/src/components/deposit/Deposit.tsx
```diff
@@ -1068,14 +1068,16 @@ const Deposit: FC<DepositProps> = ({ narrow }) => {
                   </div>
                 )}
                 {
-                  <div className="tf__flex-row-space-between-v">
+                  <div className="tf__flex-row-space-between-v test">
```
lazarmitic, line 1071: This 'test' class should be removed.

lazarmitic (changes requested):

packages/tempus-client_v2/src/components/deposit/Deposit.tsx
danieledamiani: removed

lazarmitic (approved):

PR #909, opened 2022-02-07
Load pools from all networks on the dashboard

lazarmitic:
If user wallet is not connected all pools will show, if user wallet is connected, it will show pools from network that wallet is connected to.


packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -162,7 +179,9 @@ const Wallet = () => {
           setSelectedWallet(null);
           console.error('onMetaMaskSelected', error);
           if (error instanceof UnsupportedChainIdError) {
-            requestNetworkChange();
+            // TODO - We should probably request network change to network that is currently selected in the app network selector
+            // Request user to change network to ethereum mainnet
+            requestNetworkChange(1);
```
ShawTim, line 184: should we use the `supportedChainId` here instead of the magic number

ShawTim (approved):

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
lazarmitic: This will be changed in one of the next PRs. I didn't want to make this PR any larger

danieledamiani (approved):

PR #914, opened 2022-02-08
Show Rari pool only if user already has position in it - closes #908

### LINKED ISSUE #908: Show Rari pool only if user already has position in it
Check if user has a position in Rari Pool, if user does have position, show the pool, otherwise, hide the pool from the dashboard.

danieledamiani:
closes #908


packages/tempus-client_v2/src/adapters/DashboardDataAdapter.ts
```diff
@@ -117,4 +119,25 @@ export default class DashboardDataAdapter {
...
+            return true;
+          }
+
+          return this.isUserBalancePositive(row.tempusPool.address);
+        })
+      : childRows;
+  }
+
+  private isUserBalancePositive(poolAddress: string): boolean {
+    const { principals, yields } = dynamicPoolDataState[poolAddress].poolShareBalance.get();
```
lazarmitic, line 139: Avoid reading state outside of react components, if state changes, component that is using this function will not re-render.

Read state inside react component, and pass it as a parameter to functions in services/adapters.

lazarmitic (changes requested):

packages/tempus-client_v2/src/adapters/DashboardDataAdapter.ts
danieledamiani: Refactored

PR #921, opened 2022-02-09
Show Rari pool only if user already has position in it - closes #908

### LINKED ISSUE #908: Show Rari pool only if user already has position in it

danieledamiani:
closes #908


packages/tempus-client_v2/src/components/dashboard/dashboard-manager.tsx
```diff
@@ -48,3 +69,23 @@ const DashboardManager: FC = (): JSX.Element => {
 };
 
 export default DashboardManager;
```
ShawTim, line 71: do we want to put the `export default` to the last line?

packages/tempus-client_v2/src/services/getVariableRateService.ts
```diff
@@ -27,13 +27,20 @@ const getVariableRateService = (
 
   if (signerOrProvider !== undefined && signerOrProvider !== actualSignerOrProvider) {
     actualSignerOrProviders.set(chain, signerOrProvider);
+
+    let rariVault: RariVault | null = null;
+
+    if (signerOrProvider instanceof JsonRpcSigner) {
+      rariVault = new RariVault(getProviderFromSignerOrProvider(signerOrProvider) as any);
```
ShawTim, line 34: so if `signerOrProvider ` is not an instance of `JsonRpcSigner`, rariVault will be `null`, and all rari related logic will be skipped?

danieledamiani: yes but I think the issue was solved by @lazarmitic latest PR into our Rari's SDK fork

packages/tempus-client_v2/src/components/dashboard/dashboard-manager.tsx
danieledamiani: if you like it more..

lazarmitic (approved):

lazarmitic (approved):

PR #933, opened 2022-02-09
Prevent MetaMask auto reconnect after disconnect

lazarmitic:


packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -330,7 +335,7 @@ const Wallet = () => {
           if (typeof chainId === 'string' && !supportedChainIds.includes(parseInt(chainId))) {
             // TODO - We should probably request network change to network that is currently selected in the app network selector
             // Request user to change network to ethereum mainnet
-            requestNetworkChange(1);
+            requestNetworkChange(1, false);
```
danieledamiani, line 338: what is `false`?

lazarmitic: If you hover over function name in VSCode, you will see function signature with all params and their names.

danieledamiani: when I do it, I'm seeing only one param `chainId: number`

danieledamiani: Untitled

danieledamiani: ah I was missing something

danieledamiani (approved):

PR #938, opened 2022-02-09
Keep Wallet network in sync with app network when changing network in app

lazarmitic:


packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -90,8 +90,50 @@ const Wallet = () => {
...
+                decimals: tokenDecimals,
+              },
+              rpcUrls: [rpc],
+              blockExplorerUrls: [blockExplorer],
+            },
+          ],
+        });
+
+        return true;
+      } catch (error) {
```
danieledamiani, line 124: we should log the error

danieledamiani (approved):

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
lazarmitic: fixed

PR #946, opened 2022-02-09
Add Fantom Yearn USDC pool - closes #945

### LINKED ISSUE #945: Add Fantom Yearn USDC pool

danieledamiani:
closes #945


ShawTim (approved):

packages/tempus-client_v2/src/config/config.ts
```diff
@@ -104,24 +104,24 @@ const config: Config = {
...
+        startDate: 1644432845000,
+        maturityDate: 1661500800000,
         protocol: 'yearn',
         protocolDisplayName: 'Yearn',
-        backingToken: 'DAI',
-        yieldBearingToken: 'yvDAI',
-        spotPrice: '2500',
+        backingToken: 'USDC',
+        yieldBearingToken: 'yvUSDC',
+        spotPrice: '1',
```
lazarmitic, line 120: spot price for USDC should not be 1, it's too small, we should set it to ~2500

packages/tempus-client_v2/src/config/config.ts
```diff
@@ -104,24 +104,24 @@ const config: Config = {
...
         decimalsForUI: 2,
-        showEstimatesInBackingToken: true,
+        showEstimatesInBackingToken: false,
```
lazarmitic, line 122: I think we want to show estimates in USDC here, we don't want to show estimates in Backing token only when ETH is backing token, because users can't withdraw to ETH, only to stETH.

packages/tempus-client_v2/src/config/config.ts
```diff
@@ -104,24 +104,24 @@ const config: Config = {
...
         tokenPrecision: {
-          backingToken: 18,
+          backingToken: 6,
           lpTokens: 18,
           principals: 18,
```
lazarmitic, line 126: If backing token precision is 6, I think principals and yields also have 6 precision, please check with Yuval.

lazarmitic (changes requested):

PR #1018, opened 2022-02-14
App keeps refreshing when Wallet not connected - closes #1017

### LINKED ISSUE #1017: App keeps refreshing when Wallet not connected

danieledamiani:
closes #1017

### DIFF
--- a/packages/tempus-client_v2/src/components/wallet/Wallet.tsx
+++ b/packages/tempus-client_v2/src/components/wallet/Wallet.tsx
@@ -285,7 +285,8 @@ const Wallet = () => {
       setConnecting(true);
       connect(lastSelectedWallet);
     },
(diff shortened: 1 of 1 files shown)


packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -462,7 +462,8 @@ const Wallet = () => {
         checkMetaMaskConnection();
       }
     }
-  }, [active, setWalletData, activate, setSelectedWallet, selectedChain, unsupportedNetwork]);
```
lazarmitic, line 465: I think we should keep all dependencies, except `selectedChain`, that's the only dependency that changes on every re-render.

We should remove it from all hook dependency lists in Wallet.tsx

lazarmitic (approved):

ShawTim (approved):

PR #1035, opened 2022-02-15
Fix wallet connect wrong RPC issue

lazarmitic:


packages/tempus-client_v2/src/utils/chainFromSigner.ts
```diff
@@ -0,0 +1 @@
+
```
danieledamiani, line 1: empty file?

danieledamiani (approved):

ShawTim (approved):

danieledamiani (approved):

PR #1051, opened 2022-02-17
Tempus fees block range fix for variable APR

lazarmitic:
This also updates calculation to use shares precision, so far it worked great as it is, but if we at some point have backing token with precision larger then 18 decimals, it would break.


lazarmitic:
I need to fix tests.

danieledamiani (approved):

lazarmitic:
tests fixed

ShawTim (approved):

danieledamiani (approved):

PR #1073, opened 2022-02-23
Improve unsupported network handling when fetching provider from wallet

lazarmitic:


packages/tempus-client_v2/src/utils/getProvider.ts
```diff
@@ -10,7 +10,11 @@ export default async function getProvider(
   if (userWalletSigner) {
     const chainId = await userWalletSigner.getChainId();
 
-    if (getChainNameFromId(chainId) === chain) {
+    const userWalletChain = getChainNameFromId(chainId);
+    // In case we do not support currently selected chain in user wallet this will be null
+    if (!userWalletChain) {
```
danieledamiani, line 15: To be a bit more readable it should be 
```
 if (userWalletChain === chain) {
      return userWalletSigner.provider;
 }
return getDefaultProvider(chain);
```

danieledamiani (approved):

packages/tempus-client_v2/src/utils/getProvider.ts
lazarmitic: Fixed

danieledamiani (approved):

Issue #1043, opened 2022-02-16
Approve button on withdrawal active with null value in input box


RTP2016:
Steps 
1. Go to the pool page 
2. Withdraw
3. Issue: I am able to approve 0.00 amount for withdrawal

This does not make sense to allow a user to approve zero amounts

Expected behaviour: 
DIsable the approve button when there is zero or null in the 'FROM' input boxes

<img width="639" alt="Screenshot 2022-02-16 at 09 29 05" src="https://user-images.githubusercontent.com/17698024/154236271-a5312619-9a29-4a89-a9eb-60bcba2274e4.png">

irinelT:
This issue is still reproducible.

![image](https://user-images.githubusercontent.com/98878781/156155090-d8cfbede-a817-47aa-9f7c-00b2a8fca118.png)

irinelT:
This is fixed.

PR #1077, opened 2022-02-23
Approve button on withdrawal active with null value in input box - cl…

danieledamiani:
…oses #1043


lazarmitic:
I think this logic should be part of approve button:

1. If amountToApprove is zero - disable the button.
2. If amountToApprove is less then allowance - enable approve button.

danieledamiani:
> I think this logic should be part of approve button:
> 
> 1. If amountToApprove is zero - disable the button.

not sure I understand, isn't what the PR already does?

lazarmitic:
All the logic in this PR is inside Withdraw component, and it only fixes this problem in withdraw component. Instead of implementing same logic in every component (Deposit, Withdraw, Swap, etc...) we should just add this check inside approve button itself, and this issue will be fixed in all components that use approve button.

lazarmitic (approved):

PR #1095, opened 2022-02-25
Add handler for wallet disconnect event

lazarmitic:


ShawTim (approved):

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -129,6 +129,42 @@ const Wallet = () => {
...
+  useEffect(() => {
+    if (!active) {
+      return;
+    }
+
+    const subscribeToAccountChanges = async () => {
+      const injectedConnector = new InjectedConnector({ supportedChainIds });
+      const provider = await injectedConnector.getProvider();
+
+      provider.on('accountsChanged', (accountsList: string[]) => {
```
danieledamiani, line 144: we should also unsubscribe

danieledamiani (approved):

packages/tempus-client_v2/src/components/wallet/Wallet.tsx
lazarmitic: I couldn't find method on provider that is used for unsubscribing, it seems it's automatically done as soon as wallet is disconnected/changed.

danieledamiani: is `provider.off`not available?

lazarmitic: No, I thought that something like that would be there, but I didn't find it.

PR #1104, opened 2022-02-28
refactor and update the style of sidebar

ShawTim:
update sidebar font styles


ShawTim:
for best practice we should get rid of `!important` and use what `Typography` can provide. looking forward for v3 to fix this

danieledamiani (approved):

PR #1138, opened 2022-03-08
Update dashboard maturity column UI

lazarmitic:


packages/tempus-client_v2/src/components/typography/Typography.tsx
```diff
@@ -24,9 +24,21 @@ type TypographyVariant =
...
+type TypographyColor =
+  | 'default'
+  | 'primary'
+  | 'accent'
+  | 'inverted'
+  | 'link'
+  | 'title'
+  | 'error'
+  | 'success'
+  | 'label-gray';
```
ShawTim, line 41: we may want to avoid using actual color as a value of `TypographyColor` here but i dont have a good idea.

ShawTim (approved):

packages/tempus-client_v2/src/components/typography/Typography.tsx
lazarmitic: We are going to get proper color palette for Typography from design team soon and we can add it here. Hopefully that will solve all of our issues tech dept we have here in Typography.

PR #1188, opened 2022-03-15
Automated tests

irinelT:
- Adding some BDD tests for future use.
- Adding initial automated tests.
- Setting up CI for automated tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,6 +77,7 @@ jobs:
           command: |
             xvfb-run npm run e2e || true
             ls /root/project/test-results
+            ls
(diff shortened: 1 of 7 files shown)


irinelT:
> Can we add metamask package to gitignore?

I don't think it would work without it because its not a npm package, its a Chrome extension. Do you know any way to remove it but and make it work on CI?

lazarmitic:
We can check if there is a way to install extension through a script inside CI environment.

irinelT:
> We can check if there is a way to install extension through a script inside CI environment.

Ok, I will check it out. I think that is posible.

PR #1198, opened 2022-03-17
update dashboard UI, and toggle expand/collapse through parent row

ShawTim:


packages/tempus-client_v2/src/components/dashboard/bodySection/bodyRow.tsx
```diff
@@ -1,11 +1,34 @@
-import { FC } from 'react';
+import { FC, useCallback } from 'react';
```
danieledamiani, line 1: maybe we can import `KeyboardEvent`?

danieledamiani (approved):

packages/tempus-client_v2/src/components/dashboard/bodySection/bodyRow.tsx
ShawTim: done

PR #1199, opened 2022-03-17
replace spinner with "..."

ShawTim:


packages/tempus-client_v2/src/components/wallet/Wallet.tsx
```diff
@@ -624,9 +624,13 @@ const Wallet = () => {
...
                 {/* In case there are pending transactions, show number of pending transactions */}
                 {pendingTransactions.length > 0 && (
-                  <Typography variant="h5">
-                    {getText('xxxPending', locale, { count: pendingTransactions.length })}
-                  </Typography>
+                  <>
+                    <Typography variant="h5">
+                      {getText('xxxPending', locale, { count: pendingTransactions.length })}
+                    </Typography>
+                    <Spacer size={8} />
```
ShawTim, line 631: `<Spacer>` used here for flexbox spacing, which is the purpose of it

PR #1204, opened 2022-03-18
Use utils from tempus-core-services instead of local ones

iamvukasin:


packages/tempus-client_v2/src/services/NumberUtils.ts
```diff
@@ -1,5 +1,4 @@
-import { isZeroString } from '../utils/isZeroString';
-import { reverseString } from '../utils/reverseString';
+import { isZeroString, reverseString } from 'tempus-core-services';
```
danieledamiani, line 1: this file has been moved to `tempus-core-services/src/utils` and renamed `numberUtils.ts` so you can delete it from `tempus-client_v2`

danieledamiani (commented):
a comment

danieledamiani (approved):

PR #1236, opened 2022-03-23
Add bundlewatch test

lazarmitic:


bundlewatch.config.js
```diff
@@ -0,0 +1,9 @@
+const bundlewatchConfig = {
+  files: [
+    {
+      path: 'packages/tempus-client_v2/build/static/js/*.js',
```
ShawTim, line 4: should we watch those images/css as well?

lazarmitic: Yes, I will include everything if I manage to make it work :D

PR #1254, opened 2022-03-24
introduce debouncing (300ms) for currency input

ShawTim:


packages/tempus-client_v2/src/components/currencyInput/currencyInput.tsx
```diff
@@ -36,25 +36,38 @@ const CurrencyInput: FC<CurrencyInputProps> = ({
   const { locale } = useContext(LocaleContext);
 
   const [value, setValue] = useState<string>('');
+  const [time, setTime] = useState<NodeJS.Timeout>();
```
lazarmitic, line 39: Variable `time` is not affecting component render output, so instead of using state to store it, you should use ref.

By using ref we are going to avoid unnecessary re-render every time `time` variable is changed.

lazarmitic (changes requested):

packages/tempus-client_v2/src/components/currencyInput/currencyInput.tsx
ShawTim: good catch ser! `useState` trigger rerender which is not required here.

ShawTim: fixed

danieledamiani (approved):

lazarmitic (approved):

PR #1310, opened 2022-03-29
Add color constants base

lazarmitic:


packages/tempus-client_v3/package.json
```diff
@@ -47,7 +47,8 @@
     "react-scripts": "4.0.3",
     "recharts": "^2.1.6",
     "rxjs": "^7.3.0",
-    "sass": "^1.35.1",
+    "sass": "^1.49.9",
+    "node-sass": "^7.0.1",
```
danieledamiani, line 51: do we need `node-sass`?

danieledamiani (approved):
a comment

packages/tempus-client_v3/package.json
lazarmitic: ah, I forgot to remove it. I will do it now.

PR #1313, opened 2022-03-29
Add checkbox component

iamvukasin:


packages/tempus-client_v3/src/colors.ts
```diff
@@ -1,4 +1,4 @@
-export const colors = { textPrimary: '#222222' };
+export const colors = { textPrimary: '#222222', primaryMain: '#2B6E84' };
```
danieledamiani, line 1: at some point we want to export this json file to a separate file

danieledamiani (approved):

PR #1319, opened 2022-03-29
Move atomic components to shared folder

iamvukasin:


packages/tempus-client_v3/src/components/Checkbox/Checkbox.tsx
```diff
@@ -1,4 +1,4 @@
-import Typography from '../Typography/Typography';
+import Typography from '../shared/Typography';
```
danieledamiani, line 1: I think we should have an `index.ts` at the root of `/shared` and then import like `import { Typography, SomethingElse } from `../shared`

packages/tempus-client_v3/src/index.tsx
```diff
@@ -3,7 +3,7 @@ import ReactDOM from 'react-dom';
 import isMobile from 'is-mobile';
 import reportWebVitals from './reportWebVitals';
 import Checkbox from './components/Checkbox';
-import Typography from './components/Typography/Typography';
+import Typography from './components/shared/Typography';
```
danieledamiani, line 6: see my other comment

packages/tempus-client_v3/src/components/shared/Typography/Typography.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { CSSProperties, FC } from 'react';
 import parse from 'html-react-parser';
-import { colors } from '../../colors';
+import { colors } from '../../../colors';
```
danieledamiani, line 3: `colors` may be moved in the shared folder

packages/tempus-client_v3/src/components/shared/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './Color';
+
+export { default as Icon } from './Icon';
+export * from './Icon';
+
+export { default as Typography } from './Typography';
```
danieledamiani, line 6: let's get rid of all `default`

danieledamiani (approved):

PR #1325, opened 2022-03-29
Add toggle switch component

iamvukasin:


packages/tempus-client_v3/src/components/ToggleSwitch/index.ts
```diff
@@ -0,0 +1,3 @@
+import ToggleSwitch from './ToggleSwitch';
```
danieledamiani, line 1: is the `ToggleSwitch` going to be "shared"?

packages/tempus-client_v3/src/components/shared/Shadow/index.ts
```diff
@@ -0,0 +1,10 @@
+import allShadows from './shadows.json';
+
+export const shadows = allShadows;
+
+const root = document.querySelector(':root') as HTMLElement;
+if (root) {
```
danieledamiani, line 6: this function looks pretty similar to the other we use to load up the colours...

danieledamiani (commented):
a few comments, in general lgtm

packages/tempus-client_v3/src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,39 @@
...
+import './toggle-switch.scss';
+
+interface ToggleSwitchProps {
+  checked: boolean;
+  label?: string;
+}
+
+let idCounter = 0;
+
+const ToggleSwitch = (props: ToggleSwitchProps & React.HTMLProps<HTMLInputElement>) => {
```
lazarmitic, line 12: Use FC to define props for a functional component - this way you are overriding children and some other built in props.

Let's not use ...checkBox props - define manually only what you need instead of importing all checkbox props into a component.

packages/tempus-client_v3/src/components/shared/Color/colors.json
```diff
@@ -1,4 +1,9 @@
 {
   "textPrimary": "#222222",
-  "primaryMain": "#2B6E84"
+
+  "primaryMain": "#2B6E84",
+  "primaryLight": "#129CA4",
+  "primaryDark": "#3E4E6E",
+
+  "greyLight": "#BFC0C0"
```
lazarmitic, line 8: Let's avoid using colors that have actual color value in their name - in this case 'gray' - name this color based on usage - later on when we add dark theme, we only need to update color values and names can stay the same.

packages/tempus-client_v3/src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <span className="tc__toggle-switch__switch-container">
+        <input {...checkboxProps} id={id} checked={checked} type="checkbox" />
+        <span className={`tc__toggle-switch__thumb ${checked ? 'tc__toggle-switch__thumb-active' : ''}`}></span>
+        <span
+          className={`tc__toggle-switch__background ${checked ? 'tc__toggle-switch__background-active' : ''}`}
+        ></span>
+      </span>
+      {label && (
+        <label htmlFor={id}>
+          <Typography
```
lazarmitic, line 27: Use weight property to make text bold `weight: {checked ? 'bold' : 'regular'}`, this will also help you with getting rid of !important flag in CSS.

packages/tempus-client_v3/src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,39 @@
...
+            variant="body-secondary"
+            className={`tc__toggle-switch__label ${checked ? 'tc__toggle-switch__label-selected' : ''}`}
```
lazarmitic, line 29: Instead of modifying typography with left margin - set padding on parent component. (Or use Spacer component to add space between elements inside flexbox).

lazarmitic (changes requested):

lazarmitic (approved):

PR #1335, opened 2022-03-30
base components for button and link

ShawTim:


packages/tempus-client_v3/src/components/shared/Button/index.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export interface ButtonProps {
+  children: ReactNode;
+  className?: string;
+  title?: string;
+  disabled?: boolean;
+  onClick?: () => void;
+}
+
+const Button = forwardRef((props: ButtonProps, ref: ForwardedRef<HTMLButtonElement>) => {
```
lazarmitic, line 13: Use FC to define props type, this way you don't have to specify `children` inside your interface, React will take your interface, and add any built in props under the hood.

danieledamiani (approved):

packages/tempus-client_v3/src/components/shared/Link/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export interface LinkProps {
+  children: ReactNode;
+  className?: string;
+  title?: string;
+  href?: string;
+  onClick?: () => void;
+}
+
+const Link = forwardRef((props: LinkProps, ref: ForwardedRef<HTMLAnchorElement>) => {
```
lazarmitic, line 14: Same here, user FC to define props type.

lazarmitic (changes requested):

packages/tempus-client_v3/src/components/shared/Button/index.tsx
lazarmitic: Ignore this, it does not work with memo(), seems like an issue with React typings files.

packages/tempus-client_v3/src/components/shared/Link/index.tsx
lazarmitic: Ignore this, it does not work with memo(), seems like an issue with React typings files.

PR #1355, opened 2022-04-02
SwitcherButton for network/wallet switcher

ShawTim:


packages/tempus-client_v3/src/components/shared/SwitcherButton/SwitcherButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  label?: string;
+  title?: string;
+  selected: boolean;
+  onClick: () => void;
+}
+
+const SwitcherButton: FC<SwitcherButtonProps> = props => {
+  const { logoType, label, title, selected, onClick } = props;
+
+  const weight = useMemo(() => (selected ? 'bold' : 'regular'), [selected]);
```
lazarmitic, line 21: I think we should not use `useMemo` for simple boolean checks. In this case `useMemo` is probably slowing app more then it is speeding it up.

packages/tempus-client_v3/src/components/shared/SwitcherButton/switcher-button.scss
```diff
@@ -0,0 +1,25 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  .tc__switcher-btn__label {
+    display: flex;
+    justify-content: flex-start;
+    align-items: center;
+
+    svg {
```
danieledamiani, line 17: from the `spec` file it seems `svg` is not a descendant of `.tc__switcher-btn__label`

packages/tempus-client_v3/src/components/shared/SwitcherButton/SwitcherButton.tsx
ShawTim: good comment

packages/tempus-client_v3/src/components/shared/SwitcherButton/switcher-button.scss
ShawTim: it is. there are 2 svg here. this one is for the svg with `.tc__switcher-btn__label` in specific, which is the checkmark icon

lazarmitic (approved):

danieledamiani (approved):

PR #1356, opened 2022-04-04
Add WalletButton component

lazarmitic:


lazarmitic:
CI is failing because of the breaking changes in tempus-core-services.

packages/tempus-client_v3/src/components/shared/WalletButton/WalletButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+  chain: Chain;
+  onConnect: () => void;
+  onNetworkClick: () => void;
+  onWalletClick: () => void;
+}
+
+const WalletButton: FC<WalletButtonProps> = props => {
+  const { address, balance, chain, onConnect, onNetworkClick, onWalletClick } = props;
+
+  const selectedChainTokenTicker = chainToTicker(chain);
```
danieledamiani, line 23: I think an `useMemo` would do here

packages/tempus-client_v3/src/components/shared/WalletButton/WalletButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const supportedChain = chain !== 'unsupported';
+  const shortAddress = shortenAccount(address);
```
danieledamiani, line 25: same here

packages/tempus-core-services/src/interfaces/Chain.ts
```diff
@@ -1 +1 @@
-export type Chain = 'ethereum' | 'fantom' | 'ethereum-fork';
+export type Chain = 'ethereum' | 'fantom' | 'ethereum-fork' | 'unsupported';
```
danieledamiani, line 1: are we sure this won't break any service that relies on Chain name, for example `getChainlinkFeed`?

lazarmitic: I updated getChainlinkFeed to contain empty object for unsupported network. And the UI itself should never show anything or query anything when unsupported network is selected. So my guess is this will not break anything.

danieledamiani (approved):

PR #1369, opened 2022-04-04
Add term tabs

iamvukasin:


packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.scss
```diff
@@ -0,0 +1,26 @@
...
+}
+
+.tc__term-tabs__term {
+  margin-right: 8px;
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  background-color: #ffffff;
```
danieledamiani, line 13: use var please

packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.scss
```diff
@@ -0,0 +1,26 @@
...
+  border: 1px solid #bfc0c0;
```
danieledamiani, line 14: use a colour var please

packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.scss
```diff
@@ -0,0 +1,26 @@
...
+  box-sizing: border-box;
+  border-radius: 4px;
+
+  &:last-child {
+    margin-right: 0;
+  }
+}
+
+.tc__term-tabs__term-selected {
+  background: #f3f7f9;
```
danieledamiani, line 24: use a colour var please

packages/tempus-client_v3/src/components/shared/Spacer/Spacer.tsx
```diff
@@ -0,0 +1,11 @@
+import { FC } from 'react';
+
+interface SpacerProps {
+  size: number;
+}
+
+const Spacer: FC<SpacerProps> = props => {
```
danieledamiani, line 7: it may seem redundant but a simple spec with a snapshot could save us time in the future

packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <ButtonWrapper
+      className={`tc__term-tabs__term ${selected ? 'tc__term-tabs__term-selected' : ''}`}
+      onClick={handleClick}
+    >
+      <Typography variant="body-secondary" weight="medium">
+        Term
+      </Typography>
+      <Spacer size={4} />
```
ShawTim, line 31: it will be great if we can avoid to have `Spacer` here. in this case we can handle it by CSS by setting the margin.
even if we want to have an element here in flexbox for padding, we should prefer to put the sizing in CSS instead of the javascript/typescript like:
```
.tc__spacer {
  width: 4px;
}
```

packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.spec.tsx
```diff
@@ -0,0 +1,27 @@
...
+  it('calls `onChange` when a term tab is clicked', () => {
+    const { getAllByRole } = subject({
+      dates: [new Date(2022, 0, 1), new Date(2022, 5, 1), new Date(2022, 10, 1)],
+      onChange: mockOnChange,
+    });
+
+    const buttons = getAllByRole('button');
+    fireEvent.click(buttons[1]);
+
+    expect(mockOnChange).toHaveBeenCalledTimes(1);
```
ShawTim, line 25: should check the param passed in the mock function as well

packages/tempus-client_v3/src/components/shared/Spacer/Spacer.tsx
```diff
@@ -0,0 +1,11 @@
+import { FC } from 'react';
+
+interface SpacerProps {
+  size: number;
```
lazarmitic, line 4: We should have two props here, one for width and one for height, call them `vertical` and `horizontal`, this way the component will be more flexible and we can set different spacing of each axis.

Imagine you have a flexbox row, and you want 40px distance between to elements on horizontal axis, if you add spacer with 40px size, you will get what you needed, but you will also unintentionally create a space between two separate rows vertically.

packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.spec.tsx
```diff
@@ -0,0 +1,27 @@
+import { fireEvent, render } from '@testing-library/react';
+import TermTabs, { TermTabsProps } from './TermTabs';
+
+const mockOnChange = jest.fn<Date, any>();
```
lazarmitic, line 4: Try to avoid using any.

packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <Typography variant="body-primary" type="mono">
+        <FormattedDate date={date} size="medium" separatorContrast="high" />
+      </Typography>
+    </ButtonWrapper>
+  );
+};
+
+const TermTabs: FC<TermTabsProps> = props => {
```
lazarmitic, line 39: Let's try to have one component per file.

packages/tempus-client_v3/src/components/shared/TermTabs/TermTabs.tsx
```diff
@@ -0,0 +1,59 @@
...
+      setSelectedDate(value);
+      onChange?.(selectedDate);
+    },
+    [selectedDate, onChange],
+  );
+
+  return (
+    <div className="tc__term-tabs">
+      {dates.map((termDate, index) => (
+        <TermTab date={termDate} selected={selectedDate === termDate} onClick={onClick} key={`term-tab-${index}`} />
```
lazarmitic, line 53: It's not a good idea to use only index as a key - you can termDate timestamp as well to make sure it's unique even when elements re-order, etc...

lazarmitic (approved):

PR #1370, opened 2022-04-04
Add test coverage and threshold

danieledamiani:
**NOTE: not to be approved/merged without prior team discussion

### DIFF
--- a/packages/tempus-client_v3/package.json
+++ b/packages/tempus-client_v3/package.json
@@ -99,10 +99,30 @@
       }
     ]
   },
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.{ts,tsx}",
+      "!src/**/*.stories.{ts,tsx}",
+      "!src/**/index.{ts,tsx}",
+      "!src/**/*.d.ts",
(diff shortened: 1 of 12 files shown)


ShawTim (approved):

lazarmitic (approved):

packages/tempus-client_v3/package.json
```diff
@@ -99,10 +99,36 @@
...
+    "collectCoverageFrom": [
+      "src/**/*.{ts,tsx}",
+      "!src/**/*.stories.{ts,tsx}",
+      "!src/**/index.{ts,tsx}",
+      "!src/**/*.d.ts",
+      "!src/**/reportWebVitals.ts",
+      "!src/**/serviceWorker.ts",
+      "!**/node_modules/**"
+    ],
+    "coverageReporters": [
```
lazarmitic, line 112: Based on Jest docs, these four coverage reporters are enabled by default, no need to explicitly set them here.

https://jestjs.io/docs/configuration#coveragereporters-arraystring--string-options

danieledamiani: right, removed

PR #1375, opened 2022-04-05
Add FormattedDate spec

danieledamiani:


packages/tempus-client_v3/src/components/shared/FormattedDate/FormattedDate.tsx
```diff
@@ -53,7 +53,7 @@ const FormattedDate: FC<FormattedDateProps> = props => {
   }
 
   return (
-    <span className="tc__formatted-date">
+    <span className="tc__formatted-date" data-testid="formatted-date">
```
lazarmitic, line 56: Do we need this testid attr?

ShawTim: i think we can do:
```
const { container } = subject(defaultProps);
container.querySelector('xxx');
```
not a fan to add `data-testid` becoz of unit test

danieledamiani: to target the parent container, if there's a better solution it's more than welcome :)

danieledamiani: fixed

ShawTim (approved):

PR #1384, opened 2022-04-06
text input and number input

ShawTim:


packages/tempus-client_v3/src/components/shared/Colors/colors.json
```diff
@@ -59,5 +61,12 @@
   "dropdownBackground": "#FFFFFF",
   "dropdownBorder": "#BFC0C0",
   "dropdownBackgroundHighlight": "#FFFFFF80",
-  "dropdownBackgroundSelected": "#F3F7F9"
+  "dropdownBackgroundSelected": "#F3F7F9",
+
+  "textInputBackground": "#FFFFFF",
```
lazarmitic, line 66: Add new colors to Storybook.

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,51 @@
...
+import TextInput from '../TextInput';
+import ButtonWrapper from '../ButtonWrapper';
+import Typography from '../Typography';
+
+import './NumberInput.scss';
+
+export interface NumberInputProps {
+  label?: string;
+  value?: string;
+  max?: number;
```
lazarmitic, line 11: We are working with BigNumbers most of the time, so max value should probably be string or BigNumber as well.

I would also add `precision` to the props, and limit number of decimals user can enter in the input.

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.scss
```diff
@@ -0,0 +1,23 @@
+.tc__text-input {
+  .tc__text-input__container {
+    input {
+      border-radius: 4px 0 0 4px;
```
lazarmitic, line 4: Input border is set to `none`, there is no need to set the border since it's not visible.

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.scss
```diff
@@ -0,0 +1,23 @@
+.tc__text-input {
```
lazarmitic, line 1: Instead of modifying `tc__text-input` directly, I think it would be better to add `div` wrapper around `<TextInput>` inside `<NumberInput>` component and set `className` on that `div` to be `tc__number-input`.

Inside styles here you would then do:
```
.tc__number-input {
  .tc__text-input {
   ...
  }
}
```

I think this will scope styles better and prevent styles from overriding each other.

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.spec.tsx
```diff
@@ -0,0 +1,354 @@
+import { fireEvent, render } from '@testing-library/react';
```
lazarmitic, line 1: Most of the tests here are for TextInput, I think if component A is using component B, we should not cover component B functionality in tests for component A. In this case NumberInput should only test NumberInput specific functionality, for example clicking on Max button, or trying to enter value that's not a number, etc...

This is just my opinion, we can discuss on Discord with others how we should proceed here.

packages/tempus-client_v3/src/components/shared/TextInput/TextInput.scss
```diff
@@ -0,0 +1,60 @@
...
+    border: 1px solid var(--textInputBorder);
+    border-radius: $textinput-border-radius;
+    box-sizing: border-box;
+    box-shadow: var(--downWeakShadow);
+    height: 40px;
+
+    input {
+      background: transparent;
+      border: none;
+      border-radius: $textinput-border-radius;
```
lazarmitic, line 18: Border is set to none, no need to set radius for invisible border.

packages/tempus-client_v3/src/components/shared/Colors/colors.json
ShawTim: added

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.scss
ShawTim: as discussed we still need `border-radius` to avoid overlapping with corners

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.spec.tsx
ShawTim: agree. removed most, keep remaining for snapshot purpose and to test the `NumberInput` function

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.scss
ShawTim: done

packages/tempus-client_v3/src/components/shared/NumberInput/NumberInput.tsx
```diff
@@ -18,9 +20,9 @@ export interface NumberInputProps {
 }
 
 const NumberInput: FC<NumberInputProps> = props => {
-  const { label, value, max, placeholder, caption, error, disabled, debounce, onChange } = props;
+  const { label, value, max, precision = 18, placeholder, caption, error, disabled, debounce, onChange } = props;
 
-  const onMaxClick = useCallback(() => onChange?.(`${max}`), [max, onChange]);
+  const onMaxClick = useCallback(() => onChange?.(Number(max).toFixed(precision)), [max, precision, onChange]);
```
lazarmitic, line 25: Avoid using BigNumber -> Number conversion. BigNumber can be really large and it will not fix into a Number and app will crash. We should only use BigNumber for math, and strings for UI display.

ShawTim: done

lazarmitic (approved):

PR #1402, opened 2022-04-08
Enable default-case ESLint rule

lazarmitic:


danieledamiani (approved):

packages/tempus-client_v3/src/components/shared/ActionButton/ActionButton.tsx
```diff
@@ -51,6 +51,7 @@ const ActionButton: FC<ButtonProps> = props => {
       backgroundClass += '-success';
       hoverClass += '-success';
       break;
+    default:
```
danieledamiani, line 54: this probably will reduce test coverage...

lazarmitic: Good point

lazarmitic: I compared coverage reports and it seems that this change does not reduce test coverage, in some cases it increases it (I'm not sure how). For example:
```
withIcon.tsx - with default statement set
|   33.33 |    16.66 |     100 |   33.33 | 16-29

withIcon.tsx - without default statement set
|   33.33 |        0 |     100 |   33.33 | 16-29
```

PR #1405, opened 2022-04-08
Add currency input

iamvukasin:


packages/tempus-client_v3/src/components/shared/CurrencyInput/PercentageButton.tsx
```diff
@@ -0,0 +1,26 @@
+import { FC, useCallback } from 'react';
+import { NumberUtils } from 'tempus-core-services';
+import ActionButton from '../ActionButton';
```
danieledamiani, line 3: why can't be it a `Button` component?

packages/tempus-client_v3/src/components/shared/ActionButton/ActionButton.tsx
```diff
@@ -11,8 +11,8 @@ type ButtonVariant = 'primary' | 'secondary' | 'tertiary';
 type ButtonState = 'default' | 'disabled' | 'loading' | 'success';
 type ButtonLabels = {
   default: string;
-  loading: string;
-  success: string;
+  loading?: string;
```
danieledamiani, line 14: what is the reason to change `loading` and `success` props to be optional?

packages/tempus-client_v3/src/components/shared/CurrencyInput/PercentageButton.tsx
iamvukasin: The percentage buttons should have the same style as secondary `ActionButton`s minus success and loading state.

packages/tempus-client_v3/src/components/shared/ActionButton/ActionButton.tsx
iamvukasin: `ActionButton` is being used for percentage buttons that don't have loading or success states. I marked those two as optional, so a consumer can use the button only with `labels={{ default: '...' }}` instead of `labels={{ default: '...', loading: '', success: '' }}`. It isn't a must to have that, but I think it is more clear than providing empty labels for the states that would never be used.

ShawTim: personally i think `labels={{ default: '...', loading: '', success: '' }}` makes more senses and that can be wrapped in `PercentageButton` so for the consumer it doesnt matter.
on the other hand, if there is no such state (`loading` and `success`) for `PercentageButton` we may rethink whether we want to use `ActionButton` becoz it's not but we just want to reuse the style from it. i'm open on this.

packages/tempus-client_v3/src/components/shared/CurrencyInput/CurrencyInput.scss
```diff
@@ -0,0 +1,183 @@
...
+  position: absolute;
+  top: 0;
+  left: 0;
+  margin: -#{$border-size} 0 0 -#{$border-size};
+  border: $border-size solid var(--currencyInputBorder);
+  background-color: var(--currencyInputBackground);
+  border-radius: $border-radius 0 $border-radius $border-radius;
+  box-sizing: border-box;
+
+  & > .tc__currency-input__currency-selector-item {
```
ShawTim, line 53: we dont need the `&` here

packages/tempus-client_v3/src/components/shared/CurrencyInput/CurrencyInput.scss
```diff
@@ -0,0 +1,183 @@
+$currency-selector-width: 144px;
+$currency-selector-height: 64px;
+$border-size: 1px;
+$border-radius: 8px;
+
+.tc__currency-input {
+  width: 520px;
+}
+
+.tc__currency-input__field-container {
```
ShawTim, line 10: i think all these css should placed under `.tc__currency-input` becoz all of them are the children of `.tc__currency-input`

packages/tempus-client_v3/src/components/shared/Typography/Typography.tsx
```diff
@@ -64,9 +66,11 @@ typographyVariantMap.set('body-tertiary', {
 
 const typographyColorMap = new Map<TypographyColor, string>();
 typographyColorMap.set('text-primary', colors.textPrimary);
+typographyColorMap.set('text-secondary', colors.textSecondary);
 typographyColorMap.set('primary-dark', colors.primaryDark);
 typographyColorMap.set('text-primary-inverted', colors.textPrimaryInverted);
 typographyColorMap.set('text-success', colors.textSuccess);
+typographyColorMap.set('text-error', colors.textError);
```
ShawTim, line 73: as above, we already have `text-error`

danieledamiani (approved):

ShawTim (approved):

PR #1432, opened 2022-04-12
Enable global-require ESLint rule

lazarmitic:


packages/tempus-client_v3/src/index.tsx
```diff
@@ -4,6 +4,7 @@ import { BrowserRouter } from 'react-router-dom';
 import isMobile from 'is-mobile';
 import reportWebVitals from './reportWebVitals';
 import { Typography } from './components/shared';
+import release from './release.json';
```
ShawTim, line 7: i think this will not work if you dont have a `release.json` locally

ShawTim: the `require()` soln is a conditional import. would love to see a soln to get rid of `require()` but i dont have any right now

lazarmitic: I tried to delete release.json and it works properly, because as soon as you run `npm run start` or `npm run build` release.json is created.

Only problem I see with this is that if someone opens VSCode he will see an error before running `npm run start` or `npm run build`, but this case will probably never occur.

danieledamiani (approved):

PR #1434, opened 2022-04-12
introduce TempusBigNumber

ShawTim:


packages/tempus-core-services/src/datastructures/TempusBigNumber.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (this.precision === precision) {
+      return this.value;
+    } else if (this.precision < precision) {
+      return increasePrecision(this.value, precision - this.precision);
+    } else {
+      return decreasePrecision(this.value, this.precision - precision);
+    }
+  }
+
+  toNumber(): number {
```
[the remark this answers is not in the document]
ShawTim: it's just for the completeness and actually i get error when i write test cases so ya we dont need this

packages/tempus-core-services/src/datastructures/TempusBigNumber.ts
```diff
@@ -0,0 +1,139 @@
...
+      const lastDigit = fraction.charAt(fractionDigits);
+      if (Number(lastDigit) < 5) {
+        if (fractionDigits > 0) {
+          return `${integral}.${fraction.slice(0, fractionDigits).padEnd(fractionDigits, '0')}`;
+        } else {
+          return integral;
+        }
+      } else {
+        if (fractionDigits > 0) {
+          const bignum = utils.parseUnits(`${integral}.${fraction.slice(0, fractionDigits)}`, fractionDigits);
```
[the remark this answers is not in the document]
ShawTim: just in case we need to have carry, say `999.999` will be rounded to `1000.00` so unless we write an `add` function for string to handle this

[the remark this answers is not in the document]
lazarmitic: Ah correct. Let's leave it as is. Thanks for clarification :D

packages/tempus-core-services/src/datastructures/TempusBigNumber.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  mul(multiplicand: TempusBigNumber | string | number | BigNumber): TempusBigNumber {
+    const tbignum = multiplicand instanceof TempusBigNumber ? multiplicand : new TempusBigNumber(multiplicand);
+
+    if (this.precision === tbignum.precision) {
+      return new TempusBigNumber(mul18f(this.value, tbignum.value, this.precision));
+    } else if (this.precision < tbignum.precision) {
+      const product = mul18f(
```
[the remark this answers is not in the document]
ShawTim: removed `weiMath` dependency since we dont need it here (and have a better implementation here)

packages/tempus-core-services/src/datastructures/TempusBigNumber.ts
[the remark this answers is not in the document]
ShawTim: removed

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,139 @@
...
+};
+
+export default class FixedPointDecimal {
+  readonly value: BigNumber;
+  readonly precision: number;
+
+  constructor(value: string | number | BigNumber, precision: number = 18) {
+    if (value instanceof BigNumber) {
+      this.value = value;
+    } else if (value === '.' || value === '') {
```
danieledamiani, line 20: if we pass `.` or an empty string we probably want to throw an error

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+    const precision = 10;
+    const decimal = new FixedPointDecimal(value, precision);
+
+    expect(decimal.value).toBe(value);
+    expect(decimal.precision).toBe(precision);
+    expect(decimal.toString()).toEqual('123.123');
+  });
+
+  [
+    { value: '0' },
```
danieledamiani, line 36: can we add also a field `expected` and change the expectation at line  47 to something like `expect(decimal).toEqual(expected);`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
```diff
@@ -0,0 +1,287 @@
+import { BigNumber, utils } from 'ethers';
+import FixedPointDecimal from './FixedPointDecimal';
+
+describe('FixedPointDecimal', () => {
+  it('constructor accepts number', () => {
+    const value = 123.123;
```
danieledamiani, line 6: we can add `const expected = BigNumber.from('123)`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
```diff
@@ -0,0 +1,287 @@
...
+    const precision = 10;
+    const decimal = new FixedPointDecimal(value, precision);
+
+    expect(decimal.value).toEqual(utils.parseUnits(`${value}`, precision));
```
danieledamiani, line 10: `expect(decimal.value. isEqualTo(expected)).toBeTrue()`

https://mikemcl.github.io/bignumber.js/#eq

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,139 @@
...
+      const diff = this.value.sub(increasePrecision(decimal.value, this.precision - decimal.precision));
+      return new FixedPointDecimal(diff, this.precision);
+    }
+  }
+
+  mul(multiplicand: FixedPointDecimal | string | number | BigNumber): FixedPointDecimal {
+    const decimal = multiplicand instanceof FixedPointDecimal ? multiplicand : new FixedPointDecimal(multiplicand);
+
+    if (this.precision === decimal.precision) {
+      const product = decreasePrecision(this.value.mul(decimal.value), this.precision);
```
danieledamiani, line 60: why do we `decreasePrecision` if `this.precision === decimal.precision`?

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,139 @@
...
+      return new FixedPointDecimal(quotient, decimal.precision);
+    } else {
+      const quotient = this.value
+        .mul(BigNumber.from(10).pow(this.precision))
+        .div(increasePrecision(decimal.value, this.precision - decimal.precision));
+      return new FixedPointDecimal(quotient, this.precision);
+    }
+  }
+
+  toBigNumber(precision: number = 18): BigNumber {
```
danieledamiani, line 96: the use case would be as follows?
``` 
const fpd = new FixedPointDecimal(123, 12);
const fpdBigNumber = fpd.toBigNumber(6);
```

But `fpd` contains already a precision, which is `12`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,139 @@
...
+    } else {
+      this.value = utils.parseUnits(`${value}`, precision);
+    }
+    this.precision = precision;
+  }
+
+  add(addend: FixedPointDecimal | string | number | BigNumber): FixedPointDecimal {
+    const decimal = addend instanceof FixedPointDecimal ? addend : new FixedPointDecimal(addend);
+
+    if (this.precision === decimal.precision) {
```
danieledamiani, line 31: I can see we're doing this check `if (this.precision === decimal.precision)` in each operation, maybe we can have a helper function that takes 2 BigNumbers and return them with the same precision

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done with test cases

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: we need to `decreasePrecision()` when multiplying 2 `BigNumber` becoz the `precision` is involved in the calculation which is not wanted. refer to `weiMath.mul18f()`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: this is to convert the `FixedPointDecimal` to `BigNumber` with the provided `precision`
in your case, `fpdBigNumber` is `BigNumber` of `123000000`
if you want to have a `BigNumber` with precision `12`, just use `fpd.value` which is `123000000000000`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: refactor-ed with a much cleaner soln, which adjust both `BigNumber` to higher precision for calculation

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: you can check the latest `add()` `sub()` `mul()` and `div()` and see how `toBigNumber(precision)` to be used to convert to a `BigNumber` with a specific `precision`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
ShawTim: updated

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
ShawTim: updated

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
```diff
@@ -0,0 +1,257 @@
+import { BigNumber, utils } from 'ethers';
+import FixedPointDecimal from './FixedPointDecimal';
+
+describe('FixedPointDecimal', () => {
```
danieledamiani, line 4: let's organise the tests with this structure:
```
describe('FixedPointDecimal')
   describe('constructor')
      [ // array of constructor good test cases]
      [ // array of constructor bad test cases]
  describe('add')
      [ // array of add good test cases]
      [ // array of add bad test cases]
  ...
```

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,98 @@
+import { BigNumber, utils } from 'ethers';
+
+const increasePrecision = (bignum: BigNumber, precision: number): BigNumber => {
```
danieledamiani, line 3: the name of the first argument should be `bigNum`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,98 @@
...
+      try {
+        this.value = utils.parseUnits(`${value}`, this.precision);
+      } catch (e) {
+        throw new Error(`Failed to parse ${value} when creating FixedPointDecimal`);
+      }
+    }
+  }
+
+  add(addend: FixedPointDecimal | string | number | BigNumber): FixedPointDecimal {
+    const decimal = addend instanceof FixedPointDecimal ? addend : new FixedPointDecimal(addend);
```
danieledamiani, line 30: I think we can even skip this check and so directly 
```const decimal = addend instanceof FixedPointDecimal ? addend : new FixedPointDecimal(addend);```

File: packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts [hunk 14]
```diff
@@ -0,0 +1,98 @@
...
+};
+
+const decreasePrecision = (bignum: BigNumber, precision: number): BigNumber => {
+  const divisor = BigNumber.from(BigNumber.from(10).pow(precision));
+  return bignum.div(divisor);
+};
+
+export default class FixedPointDecimal {
+  readonly value: BigNumber;
+  readonly precision: number = 18;
```
danieledamiani, line 15: since the value `18` is used twice we could use `DEFAULT_TOKEN_PRECISION` in `constants.ts`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
```diff
@@ -0,0 +1,257 @@
...
+      const decimal = new FixedPointDecimal(value);
+      const result = new FixedPointDecimal(expected);
+
+      expect(decimal.div(divisor)).toEqual(result);
+    }),
+  );
+
+  for (let i = 0; i < 10; i++) {
+    const value = Math.random() * 1000;
+    const divisor = Math.random() * 100;
```
danieledamiani, line 161: `const divisor = Math.random() * 100 + 1;` to avoid the risk of having `0`

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: that would require us to have a constructor to accept `FixedPointDecimal`, but doable.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
ShawTim: re-organized as suggested, see whether it's good?

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/FixedPointDecimal.spec.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/README.md
```diff
@@ -0,0 +1,65 @@
+# Data Structure
+### `FixedPointDecimal`
+`FixedPointDecimal` is an abstract representation of the daily use fixed point number, which underlying implementation are `ethers/BigNumber` with a `precision`.
+Instead of dealing with a pair of `ethers/BigNumber` and `precision`, we should use `FixedPointDecimal` for handling numbers and calculations.
+#### Constrcutor()
+```
+const decimal1 = new FixedPointDecimal(123, 6);
```
lazarmitic, line 7: Readme file is outdated, constructor no longer takes in precision as second argument.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
+import { BigNumber, utils } from 'ethers';
+import { DEFAULT_TOKEN_PRECISION } from '../constants';
+
+const increasePrecision = (bigNum: BigNumber, precision: number): BigNumber => {
```
lazarmitic, line 4: Let's rename second argument to something like `increment`.

We are not setting specific precision for `bigNum` we are increasing it.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+  const multiplicand = BigNumber.from(BigNumber.from(10).pow(precision));
```
lazarmitic, line 5: I think you have one extra `BigNumber.from()` here. Method `pow` already returns an instance of `BigNumber`.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+  return bigNum.mul(multiplicand);
+};
+
+const decreasePrecision = (bigNum: BigNumber, precision: number): BigNumber => {
```
lazarmitic, line 9: I think we should rename second argument to `decrement`.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+  const divisor = BigNumber.from(BigNumber.from(10).pow(precision));
```
lazarmitic, line 10: One extra `BigNumber.from()` here.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+  return bigNum.div(divisor);
+};
+
+export default class FixedPointDecimal {
+  readonly value: BigNumber;
+  readonly precision: number = DEFAULT_TOKEN_PRECISION;
+
+  constructor(value: string | number | BigNumber | FixedPointDecimal) {
+    if (value instanceof FixedPointDecimal) {
+      this.value = value.value;
```
lazarmitic, line 20: I think we should not assign as a reference here. Reason is:

Let's say I have `FixedPointDecimal` called `A`:

1. I want to create new `FixedPointDecimal` called `B`.
2. I do that by passing `A` to `B`'s constructor.
3. Now `B`'s value object is same as `A`'s value object.
4. This means that if I change `A`'s value, `B`'s value will change as well because they are pointing to same object in memory.

Solution is to create new BigNumber for `B`'s value using `A`'s value.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+    } else if (value instanceof BigNumber) {
+      this.value = value;
```
lazarmitic, line 22: Same here, clone value instead of assigning it.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+    } else {
+      try {
+        this.value = utils.parseUnits(`${value}`, this.precision);
+      } catch (e) {
+        throw new Error(`Failed to parse ${value} when creating FixedPointDecimal`);
+      }
+    }
+  }
+
+  add(addend: FixedPointDecimal | string | number | BigNumber): FixedPointDecimal {
```
lazarmitic, line 32: We should create type alias `type Numberish = FixedPointDecimal | string | number | BigNumber` and use it instead of always specifying all four types. It will make code a lot easier to read.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+  div(divisor: FixedPointDecimal | string | number | BigNumber): FixedPointDecimal {
+    const decimal = new FixedPointDecimal(divisor);
+    const quotient = increasePrecision(this.value, this.precision).div(decimal.value);
+
+    return new FixedPointDecimal(quotient);
+  }
+
+  toBigNumber(precision: number = DEFAULT_TOKEN_PRECISION): BigNumber {
+    if (this.precision === precision) {
+      return this.value;
```
lazarmitic, line 60: Create new instance instead of returning a reference, similar reasoning I provided for constructor.

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (Number(lastDigit) < 5) {
+      return this.toTruncated(fractionDigits);
+    } else {
+      if (fractionDigits > 0) {
+        const bigNum = utils.parseUnits(`${integral}.${fraction.slice(0, fractionDigits)}`, fractionDigits);
+        const str = utils.formatUnits(bigNum.add(1), fractionDigits);
+        const [outputIntegral, outputFraction] = str.split('.');
+        return `${outputIntegral}.${outputFraction.padEnd(fractionDigits, '0')}`;
+      } else {
+        return BigNumber.from(integral).add(1).toString();
```
lazarmitic, line 86: Note that BigNumber.toString() and this.toString() do not return same thing.

ShawTim: yes, we use `BigNumber.toString()` here. `this.toString()` will return the wrong result

packages/tempus-core-services/src/datastructures/README.md
ShawTim: updated

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: renamed

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: renamed

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: `BigNumber` is supposed to be immutable so the scenario doesnt apply here, but i do get your point.
use `BigNumber.from()` for another instance here

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done as above

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done

packages/tempus-core-services/src/datastructures/FixedPointDecimal.ts
ShawTim: done as above

danieledamiani (approved):

lazarmitic (approved):

PR #1439, opened 2022-04-13
Setup e2e tests

irinelT:
# Set up E2E on CI.

What is done:

- Created 9 automated tests.
- All tests are iterable.
- Run Playwright container on CI.
- Store keys on CI.
- Load MetaMask to CI.
- Split tests and execute in parallel.

What needs to be done:

- MetaMask login is flaky and need to be fixed or run 2 times in order to work.
- Reports from parallel execution overwrites each other.
- More coverage is needed.
- App URL needs to be local to CI.
- Switch from .js to .ts.
- Use other reporters instead of junit.

GitHub Admin:
- [Rename checks](https://circleci.com/docs/2.0/enable-checks/#troubleshooting-github-checks-waiting-for-status-in-github) required from "test" to "unit-test" (&do the same on CI).
- Add additional checks.


lazarmitic:
We should probably rename test job to unit-test because we now have e2e-test job as well.

irinelT:
> We should probably rename test job to unit-test because we now have e2e-test job as well.

I tried that but failed because only GitHub admin can rename the Required jobs and `test` is one of those.

.circleci/config.yml
```diff
@@ -59,13 +59,70 @@ jobs:
...
             - ./
 
+  metamask:
+    docker:
+      - image: mcr.microsoft.com/playwright:focal
+        environment:
+          METAMASK_RECOVERY_PHRASE: $METAMASK_RECOVERY_PHRASE
+          METAMASK_PASSWORD: $METAMASK_PASSWORD
+          METAMASK_ACCOUNT_ETH_FORK: $METAMASK_ACCOUNT_ETH_FORK
+          METAMASK_ACCOUNT_FANTOM: $METAMASK_ACCOUNT_FANTOM
```
lazarmitic, line 76: Are we going to run e2e tests on Fantom using real tokens (money)?

I think I can spin up Fantom fork as well pretty easily (it's identical to Ethereum fork I just need to change some parameters) - let me know if this is something you need and I will do it.

lazarmitic:
If `modules`, `tests` and `utility` directories are all part of integration testing, can we put all of those into a single root directory called 'integration-tests'. I just want to keep root of the repository as clean as possible :D

.circleci/config.yml
irinelT: Good question. That is up for discussion. In my opinion using both forked and real networks is best. We can have thousands of tests on forked networks and do just 2 basic tests (deposit, withdraw) using real tokens on real network (Fantom).

irinelT:
>

irinelT:
> If `modules`, `tests` and `utility` directories are all part of integration testing, can we put all of those into a single root directory called 'integration-tests'. I just want to keep root of the repository as clean as possible :D

I agree. The plan is to do this in next PR, but I can do it now if you think its urgent.

lazarmitic:
We can do it in next RP as well :D

lazarmitic (approved):

lazarmitic:
Make sure to Squash and Merge when merging PR

PR #1444, opened 2022-04-14
Support single-choice dropdown

iamvukasin:
* `ButtonWrapper` is not used as the top-level element for `Dropdown` to prevent a nested `button`-in-`button` structure when using `DropdownSelectableItem`
* `Dropdown` should be used with `DropdownCheckboxItem`
* `DropdownSelector` should be used with `DropdownSelectableItem`


packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItem.tsx
```diff
@@ -0,0 +1,56 @@
+import { Children, FC, isValidElement, ReactElement, useCallback, useState } from 'react';
+import ButtonWrapper from '../ButtonWrapper';
+import Typography from '../Typography';
+import { DropdownSelectableItemIconProps } from './DropdownSelectableItemIcon';
+import './DropdownItem.scss';
+
+export interface DropdownSelectableItemProps {
+  label: string;
+  value?: any;
```
lazarmitic, line 9: Please avoid using any - we will probably be fine if we use 'string' as value.

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItemIcon.tsx
```diff
@@ -0,0 +1,20 @@
+import { FC } from 'react';
+import Icon, { IconType } from '../Icon';
+
+export interface DropdownSelectableItemIconProps {
+  icon: IconType;
+  // eslint-disable-next-line react/no-unused-prop-types
+  value?: any;
```
lazarmitic, line 7: If we are not currently using prop, please remove it, we can add it later if we need it.

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItem.tsx
lazarmitic: If we want to be flexible we can also use generics. This was consumer of the component can define value type.

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItem.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const alreadySelectedIndex = childrenValues ? childrenValues.findIndex(val => val === selectedValue) : null;
+
+  const [selectedIndex, setSelectedIndex] = useState<number>(
+    alreadySelectedIndex !== null && alreadySelectedIndex > -1 ? alreadySelectedIndex : 0,
+  );
+
+  const resolvedValue = childrenValues && selectedIndex !== null ? childrenValues[selectedIndex] : value;
+  const selected = selectedValue === resolvedValue;
+
+  const handleClick = useCallback(() => {
```
lazarmitic, line 29: This seems really complicated, if we want to select an item from dropdown, why not just set currently selected item ID as active, instead of doing all of this math with indexes?

Or if we don't want to use ID, we can probably use value, because I don't see anyone using same value for multiple dropdown items, usually each value is different.

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItem.tsx
```diff
@@ -0,0 +1,56 @@
+import { Children, FC, isValidElement, ReactElement, useCallback, useState } from 'react';
+import ButtonWrapper from '../ButtonWrapper';
+import Typography from '../Typography';
+import { DropdownSelectableItemIconProps } from './DropdownSelectableItemIcon';
+import './DropdownItem.scss';
+
+export interface DropdownSelectableItemProps {
+  label: string;
+  value?: any;
+  selectedValue?: any;
```
lazarmitic, line 10: What's the difference between value and selectedValue?

iamvukasin: The `value` is the item's value and `selectedValue` is what the parent dropdown passes as its currently selected value. The reason I'm working like that instead of just passing if the item is selected or not is that an item can have multiple "sub-values" (for example, an item that sorts pools alphabetically can have two directions: A-Z and Z-A. In that case, the item must be selected if any of those two filtering options is selected).

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItemIcon.tsx
iamvukasin: The prop is not used for rendering, but for calculating if the item that contains the icon is selected or not. The compiler just sees it as if it is not being used ([it is](https://github.com/tempus-finance/tempus-app/pull/1444/files/f366dcc7e4084c66284331706d9efa4bc26f2fea#diff-7f7603346e3098c49d510319d4713d17db0b936a3216d267164a7de5dfe13bfeR17), but I understand why it can't match it).

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItem.tsx
iamvukasin: I agree it looks complicated (two terms with `selected` in their names that mean different things), so I will update it with more appropriate names. There are:
1. `selected` that must be `true` if it is selected or any of its sub-states (A-Z or Z-A direction, for example).
2. `selectedIndex` should be called `selectedIconIndex` (and other connected variables should be renamed too) as it just marks which icon is currently visible.

packages/tempus-client_v3/src/components/shared/Dropdown/Dropdown.tsx
```diff
@@ -25,9 +25,11 @@ const Dropdown: FC<DropdownProps> = props => {
   }, []);
 
   return (
-    <ButtonWrapper className="tc__dropdown" onClick={onClick} selected={open}>
-      <Typography variant="body-secondary">{label}</Typography>
-      <Icon type={open ? 'up-chevron' : 'down-chevron'} size={12} />
+    <div className="tc__dropdown">
+      <ButtonWrapper onClick={onClick} selected={open}>
+        <Typography variant="body-secondary">{label}</Typography>
+        <Icon type={open ? 'up-chevron' : 'down-chevron'} size={12} />
```
danieledamiani, line 31: `size="tiny"`

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelector.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onSelect?: (value: any) => void;
+  children?: ReactElement<DropdownSelectableItemProps> | ReactElement<DropdownSelectableItemProps>[];
+}
+
+const DropdownSelector: FC<DropdownSelectorProps> = props => {
+  const { label, popupTitle, selectedValue, onSelect, children } = props;
+
+  return (
+    <Dropdown label={label} popupTitle={popupTitle}>
+      {Children.map(children, child =>
```
danieledamiani, line 18: Why not doing `{...props.children}`? If you have 100 children you don't want to attach `onSelect` to each of them, but through `event bubbling` (https://www.sitepoint.com/event-bubbling-javascript/) `Dropdown` will receive the event and from there you can call `onSelect`

packages/tempus-client_v3/src/components/shared/Dropdown/DropdownSelectableItem.tsx
lazarmitic: I was thinking about this, and maybe we can just remove `DropdownSelectableItemIcon`, add icon prop to `DropdownSelectableItem` `icon: IconType`, and then the parent component, for example dashboard, will tell dropdown which icon to render. We will probably have a state in Dashboard that keeps current sort order. Using this state in the dashboard we can tell dropdown item which icon to render with 'iconType={sortOrder === 'ascending' : 'up-arrow': 'down-arrow'}.

iamvukasin: If I understood correctly, the flow would be `Dashboard` > `DropdownSelector` > `DropdownSelectableItem` and only the selected item would have the propagated `iconType` to the right?

lazarmitic: Yes. Only selected item will show the icon.

lazarmitic (approved):

PR #1446, opened 2022-04-14
Scaffold page

danieledamiani:


packages/tempus-client_v3/src/components/PageNavigation/PageNavigation.scss
```diff
@@ -2,6 +2,7 @@
   position: relative;
   padding: 10px 20px;
   border: none;
+  margin-bottom: -1px;
```
iamvukasin, line 5: I think it is better to set `box-sizing: border-box` for tabs and change `.tc__page-navigation__background::before`'s `top` and `bottom` to 0. Plus, those SVGs must be cut better to have a seamless shape match.

packages/tempus-client_v3/src/components/App/App.scss
```diff
@@ -0,0 +1,48 @@
...
+  max-width: 1360px;
+  width: 100%;
+}
+
+.tc__app__body {
+  flex: 1 1 100%;
+  padding: 30px;
+  box-sizing: border-box;
+  width: 100%;
+  max-width: 1360px;
```
lazarmitic, line 45: It seems `1360px` is used in few places, can we create a variable for it.

packages/tempus-client_v3/src/components/PageNavigation/PageNavigation.scss
danieledamiani: Checked with @Olafkuusik and some css were fixed, the SVG should be fine now

packages/tempus-client_v3/src/components/App/App.scss
danieledamiani: refactored

lazarmitic (approved):

PR #1458, opened 2022-04-19
Fix vault statistics services

danieledamiani:


packages/tempus-core-services/src/services/VariableRateService.ts
```diff
@@ -150,7 +150,9 @@ export class VariableRateService {
     const sortedEvents = await this.getSwapAndPoolBalanceChangedEvents(poolConfig, fetchEventsFromBlock);
 
     // Fetch current pool balance
-    const { principals, yields } = await this.getPoolTokens(poolConfig.poolId, principalsAddress, yieldsAddress);
+    const poolTokens = await this.getPoolTokens(poolConfig.poolId, principalsAddress, yieldsAddress);
+    let { principals } = poolTokens;
```
ShawTim, line 154: we dont need `let` here? it should be a `const`

packages/tempus-core-services/src/services/VaultService.test.ts
```diff
@@ -78,7 +78,7 @@ describe('VaultService', () => {
     });
   });
 
-  describe('getSwapEvents()', () => {
+  describe.skip('invalid getSwapEvents()', () => {
```
ShawTim, line 81: is it intended to have `skip()` here?

ShawTim (approved):
approved with comments

PR #1486, opened 2022-04-20
DecimalUtil which is a Decimal version of NumberUtil, and extend Decimal functionality

ShawTim:


packages/tempus-core-services/src/datastructures/Decimal.ts
```diff
@@ -76,16 +104,40 @@ export default class Decimal {
     const [integral, fraction] = str.split('.');
 
     const lastDigit = fraction.charAt(fractionDigits);
-    if (Number(lastDigit) < 5) {
-      return this.toTruncated(fractionDigits);
+
+    if (this.gte(0)) {
+      // positive number
```
danieledamiani, line 109: `gte` is non-negative

packages/tempus-core-services/src/datastructures/Decimal.ts
```diff
@@ -76,16 +104,40 @@ export default class Decimal {
...
-        const str = utils.formatUnits(bigNum.add(1), fractionDigits);
-        const [outputIntegral, outputFraction] = str.split('.');
-        return `${outputIntegral}.${outputFraction.padEnd(fractionDigits, '0')}`;
+      // negative number
+      if (Number(lastDigit) <= 5) {
+        // no need to be rounded
+        return this.toTruncated(fractionDigits);
       } else {
-        return BigNumber.from(integral).add(1).toString();
+        if (fractionDigits > 0) {
```
danieledamiani, line 131: how do we reach this branch?

danieledamiani (approved):

packages/tempus-core-services/src/utils/DecimalUtils.ts
```diff
@@ -0,0 +1,51 @@
+import Decimal, { Numberish } from '../datastructures/Decimal';
+import { reverseString } from './reverseString';
+
+const MULTIPLIER_LOOKUP = [
```
danieledamiani, line 4: maybe a map would be more efficient

packages/tempus-core-services/src/utils/DecimalUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+  { value: '1000000', symbol: 'M' },
+  { value: '1000', symbol: 'k' },
+];
+
+export default class DecimalUtils {
+  // round for multiplier (e.g. 9876 -> 9.88k), trauncate for fraction (e.g. 9.876 -> 9.87)
+  static formatWithMultiplier(value: Numberish, fractionDigits: number = 0): string {
+    const decimal = new Decimal(value);
+    const decimalAbs = decimal.abs();
+    const multiplier = MULTIPLIER_LOOKUP.find(item => decimalAbs.gte(item.value));
```
danieledamiani, line 18: if we use a map we can do
```
const decimalAbsLength = decimal.toString().length;
const multiplier = MULTIPLIER_LOOKUP[decimalAbsLength];
```

packages/tempus-core-services/src/datastructures/Decimal.ts
ShawTim: updated comment

packages/tempus-core-services/src/utils/DecimalUtils.ts
ShawTim: updated as suggested

packages/tempus-core-services/src/utils/DecimalUtils.ts
ShawTim: we still need `MULTIPLIER_LOOKUP.find()` but updated as suggested to check the string length instead

lazarmitic (approved):

PR #1494, opened 2022-04-22
Add Modal component

lazarmitic:


packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export interface ModalProps {
+  open: boolean;
+  onClose: () => void;
+}
+
+const Modal: FC<ModalProps> = props => {
+  const { open, onClose, children } = props;
+
+  const onBackdropClick = (event: React.MouseEvent<HTMLDivElement>) => {
```
ShawTim, line 15: should we use `useCallback` here such that when it passed to the components it will keep the same reference

packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
```diff
@@ -0,0 +1,43 @@
...
+    event.stopPropagation();
+
+    onClose();
+  };
+
+  const onCloseButtonClick = () => {
```
ShawTim, line 21: same as above

ShawTim (approved):

packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
lazarmitic: I added `useCallback` to `onCloseButtonClick()`.

I didn't add it on `onBackdropClick()` because it's not being passed down to child elements.

packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
lazarmitic: Added `useCallback` here.

packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
ShawTim: in this case, do we want to add `React.memo` for `Modal`?

lazarmitic: Yeah, I will add `React.memo()`

lazarmitic: Fixed

PR #1496, opened 2022-04-25
Add app header

lazarmitic:


packages/tempus-client_v3/src/components/shared/Logo/ProtocolTempus.tsx
```diff
@@ -0,0 +1,42 @@
+import { FC } from 'react';
+import LogoProps from './LogoProps';
+import withLogo from './withLogo';
+
+const TokenDAI: FC<LogoProps> = () => (
```
ShawTim, line 5: `TokenDAI` should be `ProtocolTempus`, minor though

ShawTim:
i'm not sure whether we should put the `Tempus Logo` under `shared/Logo` becoz it's not in 1:1 ratio. the `ProtocolTempus` should be expected to be used as a token icons like others but it gets a 120x30 size.
i think we should just put this SVG under `Navbar` folder since `Navbar` is the only consumer.
wdyt?

lazarmitic:
Makes sense, I will move Tempus logo under Navbar directory.

lazarmitic:
Fixed

ShawTim (approved):

PR #1509, opened 2022-04-27
Add nav subheader

iamvukasin:


packages/tempus-client_v3/src/components/MarketsSubheader/MarketsSubheader.tsx
```diff
@@ -0,0 +1,126 @@
...
+    (pool: PoolType) => {
+      setPoolType(pool);
+      onPoolTypeChange?.(pool);
+    },
+    [onPoolTypeChange],
+  );
+
+  const handleViewTypeChange = useCallback(
+    (iconType: IconVariant) => {
+      let viewType: ViewType;
```
ShawTim, line 45: in general i'd try to avoid `let` but it's my own preference only.
maybe we can consider this if the logic branches is not complicated:
```
(iconType: IconVariant) => {
  if (iconType === 'grid-view') {
    onViewTypeChange?.('grid');
  } else  if (iconType === 'list-view') {
    onViewTypeChange?.('list');
  }
}
```
not a problem though

packages/tempus-client_v3/src/components/PortfolioSubheader/PortfolioSubheader.spec.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <BrowserRouter>
+      <PortfolioSubheader {...props} />
+    </BrowserRouter>,
+  );
+
+describe('PortfolioSubheader', () => {
+  it('renders a navigation subheader portfolio tabs', () => {
+    const { container } = subject({});
+
+    expect(container).not.toBeNull();
```
ShawTim, line 18: maybe you want to match the snapshot of the `container` as well

ShawTim (approved):

PR #1519, opened 2022-04-29
Add TVL component

danieledamiani:


packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -0,0 +1,75 @@
+import { BigNumber } from 'ethers';
```
lazarmitic, line 1: Use Decimal instead - we should avoid using BigNumber directly.

packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const useTvlData = (): [BigNumber] => {
+  const [tvlData, setTvlData] = useState<BigNumber>(ZERO);
+
+  useEffect(() => {
+    const statisticsService = getServices('ethereum')?.StatisticsService;
+    let tvlInterval$: any;
+
+    if (statisticsService) {
+      tvlInterval$ = interval(10000)
```
lazarmitic, line 28: Let's put fetch intervals in constants/config file.

packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -0,0 +1,75 @@
...
+        .pipe(
+          startWith(0),
+          switchMap(() => getConfig()),
+          map<Config, TempusPool[]>((config: Config) => mapConfigToPools(config)),
+          mergeMap((tempusPools: TempusPool[]) =>
+            zip<BigNumber[]>(
+              tempusPools.map((tempusPool: TempusPool) => {
+                console.log('tempusPool', tempusPool);
```
lazarmitic, line 36: Remove console logs.

packages/tempus-core-services/src/services/getServices.ts
```diff
@@ -0,0 +1,57 @@
...
+
+type Service =
+  | 'TempusPoolService'
+  | 'TempusControllerService'
+  | 'StatisticsService'
+  | 'VaultService'
+  | 'VariableRateService'
+  | 'StorageService';
+
+type ServiceMap = { [key in Service]?: any };
```
lazarmitic, line 18: Let's avoid using any - maybe we can create base class for services that services extend from, and use that base class here as a type.

packages/tempus-client_v3/src/config/getConfig.ts
```diff
@@ -35,5 +35,31 @@ export const getConfig = async (): Promise<Config> => {
     }
   }
 
+  if (config) {
+    config = {
+      ...config,
+      ethereum: {
+        ...config.ethereum,
+        privateNetworkUrl: String(process.env.REACT_APP_ETHEREUM_RPC),
+        alchemyKey: String(process.env.REACT_APP_MAINNET_ALCHEMY_KEY),
```
lazarmitic, line 44: I think we can remove alchemyKey from config. `REACT_APP_ETHEREUM_RPC` contains full RPC url for network.

packages/tempus-client_v3/src/config/getConfig.ts
```diff
@@ -35,5 +35,31 @@ export const getConfig = async (): Promise<Config> => {
...
+      },
+      fantom: {
+        ...config.fantom,
+        privateNetworkUrl: String(process.env.REACT_APP_FANTOM_RPC),
+      },
+      'ethereum-fork': {
+        ...config['ethereum-fork'],
+        privateNetworkUrl: String(process.env.REACT_APP_ETHEREUM_RPC),
+        alchemyKey: String(process.env.REACT_APP_MAINNET_ALCHEMY_KEY),
```
lazarmitic, line 53: Same here.

ShawTim:
https://react-rxjs.org/docs/api/core/bind/
if we are going to use `rxjs` we may want to take a look on `react-rxjs`, which make things in react way as a hook

ShawTim:
in general i support the idea of `useTvlData` hook. the implementation of the hook doesnt matter much to me.
so we can have similar patterns. whenever we need to have some read-only data from anywhere (chain, backend, some servers, etc), we can implement a hook `useXXXData` to retrieve these read-only raw data
another question will be, those are raw data, do we need to restructure them? if yes is it another class or in the same hook
depends on the consumers the structure required maybe slightly different

packages/tempus-client_v3/src/hooks/useTvlData.ts
danieledamiani: fixed

packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -0,0 +1,75 @@
...
+  });
+  return pools;
+};
+
+export const useTvlData = (): [BigNumber] => {
+  const [tvlData, setTvlData] = useState<BigNumber>(ZERO);
+
+  useEffect(() => {
+    const statisticsService = getServices('ethereum')?.StatisticsService;
+    let tvlInterval$: any;
```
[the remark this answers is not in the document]
danieledamiani: fixed

packages/tempus-client_v3/src/hooks/useTvlData.ts
danieledamiani: fixed temporarily, we may want to define a specific interval for this specific hook

packages/tempus-client_v3/src/hooks/useTvlData.ts
danieledamiani: done

packages/tempus-core-services/src/services/getServices.ts
danieledamiani: not easy to do it now as our services don't have a common interface

packages/tempus-client_v3/src/config/getConfig.ts
danieledamiani: it seems it's used somewhere, I had to put it back

packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -0,0 +1,84 @@
...
+    config[chain].tempusPools.forEach((pool: TempusPool) => {
+      const extendedPool: ExtendedTempusPool = { ...pool, chain: chain as Chain };
+      pools.push(extendedPool);
+    });
+  });
+
+  return pools;
+};
+
+export const useTvlData = (): [Decimal] => {
```
ShawTim, line 37: is it necessary to return `[Decimal]` instead of `Decimal`?

ShawTim (approved):

packages/tempus-client_v3/src/hooks/useTvlData.ts
danieledamiani: it's for consistency 
```
const [tvl] = useTvlData()
```

PR #1531, opened 2022-05-03
Use `TooltipWrapper` for `FeeTooltip`

iamvukasin:


packages/tempus-client_v3/src/components/FeeTooltip/FeeTooltip.tsx
```diff
@@ -65,8 +63,11 @@ const FeeTooltip: FC<FeeTooltipProps> = props => {
...
-    </Tooltip>
+    ),
+    [fees.deposit, fees.earlyRedemption, fees.redemption, fees.swap],
   );
+
+  return <TooltipWrapper tooltipContent={tooltipContent}>{children}</TooltipWrapper>;
 };
 
-export default memo(FeeTooltip);
+export default memo(FeeTooltip) as FC<FeeTooltipProps>;
```
ShawTim, line 73: recently i find that typescript doesnt work quite well (randomly) for `memo` that requires `children`.
and since `children` will not be memories anyway we can take away `memo` here.

ShawTim (approved):

PR #1533, opened 2022-05-03
add bottom-center for Tooltip, and introduce InfoTooltip

ShawTim:
image

image


packages/tempus-client_v3/src/components/shared/Tooltip/Tooltip.scss
```diff
@@ -3,6 +3,7 @@
 $arrowWidth: 20px;
 $arrowEdgeOffset: 24px;
 $arrowBorderOffset: #{math.sqrt(2)}px;
+$arrowBorderWidth: 2px;
```
iamvukasin, line 6: You should combine those two variables into one and use `sqrt(2)px` instead of `2px`. See the difference between `2px`:
Screen Shot 2022-05-04 at 09 32 02
vs `sqrt(2)px`:
Screen Shot 2022-05-04 at 09 32 58

iamvukasin: Technically, they're the same thing called differently

iamvukasin (approved):
Just update `Tooltip.stories.tsx` with `bottom-center` and that's all

ShawTim:
> Just update `Tooltip.stories.tsx` with `bottom-center` and that's all

updated

PR #1541, opened 2022-05-04
Refactor `totalValueLockedUSD` method

danieledamiani:


packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -25,14 +26,26 @@ const tvlData$: Observable<Decimal> = interval(TVL_POLLING_INTERVAL_IN_MS).pipe(
...
+            address,
+            backingToken,
+          ),
+        ),
       ),
     ),
   ),
-  map((poolLockedValues: Decimal[]) => poolLockedValues.reduce((memo: Decimal, current: Decimal) => memo.add(current))),
+  map<BigNumber[], Decimal[]>((poolLockedValuesBn: BigNumber[]) =>
+    poolLockedValuesBn.map((value: BigNumber) => new Decimal(value)),
```
ShawTim, line 43: should we move this to the `core-services`?
in this way on on hooks layer we dont need to concern about `BigNumber` and the `precision`, and we dont even need `BigNumber` under `client_v3` but just to use `Decimal` instead.

ShawTim (approved):

PR #1553, opened 2022-05-06
Add deposit modal

iamvukasin:


packages/tempus-client_v3/src/components/CurrencyInputModal/CurrencyInputModal.tsx
```diff
@@ -26,7 +26,9 @@ interface CurrencyInputModalProps extends ModalProps {
   chainConfig?: ChainConfig;
   infoRows: ReactNode;
   actionButtonLabels: ActionButtonLabels;
-  onActionButtonClick: () => string;
+  actionButtonState?: ActionButtonState;
+  onAmountChange?: (amount: Decimal) => void;
+  onActionButtonClick: (amount: Decimal) => string;
```
ShawTim, line 31: i think we should change `onActionButtonClick` to a better name becoz just checking the name i have no ideas why it's accepting `amount: Decimal` as param and why it's returning a string

packages/tempus-client_v3/src/components/CurrencyInputModal/CurrencyInputModal.tsx
```diff
@@ -44,30 +46,33 @@ const CurrencyInputModal: FC<CurrencyInputModalProps> = props => {
...
   const amountDecimal = useMemo(() => new Decimal(amount || 0), [amount]);
 
-  const disabledInput = transactionState !== 'default';
+  const disabledInput = actionButtonState !== 'default';
   const insufficientBalance = amountDecimal.gt(balance);
 
+  const handleAmountChange = useCallback(
+    (value: string) => {
+      setAmount(value);
+      onAmountChange?.(new Decimal(value || 0));
```
ShawTim, line 66: if `value` === `,` or `.` it's possible to throw an error here. you can use `Decimal.parse(value, 0)`

ShawTim (approved):

PR #1557, opened 2022-05-06
useLocale hook for the locale state

ShawTim:


packages/tempus-client_v3/src/hooks/useLocale.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { renderHook } from '@testing-library/react-hooks';
+import { act } from 'react-dom/test-utils';
+import { SUPPORTED_LOCALES, SUPPORTED_LOCALE_NAMES, useLocale } from './useLocale';
+
+describe('useLocale', () => {
+  it('by default it return en', () => {
```
danieledamiani, line 6: better to put quotes around `'en'`

danieledamiani (approved):

iamvukasin (approved):

packages/tempus-client_v3/src/hooks/useLocale.spec.ts
ShawTim: fixed

PR #1576, opened 2022-05-10
Add wallet balances hook

lazarmitic:


packages/tempus-client_v3/src/components/Wallet/Wallet.tsx
```diff
@@ -119,12 +122,21 @@ const Wallet: FC = () => {
...
   }, [wallet]);
 
+  const balance = useMemo(() => {
+    // TODO - Once we add state for currently selected chain, use it to get token balance for selected chain
+    const tokenBalance = walletBalances[`ethereum-${ZERO_ETH_ADDRESS}`];
+    if (!tokenBalance) {
+      return null;
+    }
+
+    return DecimalUtils.formatToCurrency(tokenBalance, 2);
```
iamvukasin, line 132: Just a reminder (maybe add some `TODO` comment here): use `decimalsForUI` instead of hardcoded `2`

lazarmitic: Ah, you are correct! Thanks for the heads up.

packages/tempus-client_v3/src/components/Wallet/Wallet.tsx
```diff
@@ -119,12 +122,21 @@ const Wallet: FC = () => {
     return wallet.accounts[0].address;
   }, [wallet]);
 
+  const balance = useMemo(() => {
+    // TODO - Once we add state for currently selected chain, use it to get token balance for selected chain
+    const tokenBalance = walletBalances[`ethereum-${ZERO_ETH_ADDRESS}`];
+    if (!tokenBalance) {
```
ShawTim, line 128: it's possible to be `0` besides `null` or `undefined`? if it's `0` i think we should continue

lazarmitic: It's an instance of Decimal, so it's either Decimal (value=0), or undefined.

ShawTim: you are right, the balance is `Decimal | undefined` here

packages/tempus-client_v3/src/config/ConfigManager.ts
```diff
@@ -34,7 +34,45 @@ class ConfigManager {
...
+        );
+      } else {
+        console.warn(`Pool ${pool.address} does not contain chain in it's config!`);
+      }
+    });
+
+    return result;
+  }
+
+  async retrieveConfig(): Promise<boolean> {
```
danieledamiani, line 75: why is not `private` anymore?

packages/tempus-client_v3/src/config/ConfigManager.ts
```diff
@@ -34,7 +34,45 @@ class ConfigManager {
     return pools;
   }
 
-  private async retrieveConfig(): Promise<boolean> {
+  getTokenList(): { chain: Chain; address: string }[] {
```
danieledamiani, line 37: we could store the `tokenList` in a private variable, `getTokenList` returns that variable and another private `retrieveTokenList` could be invoked after `this.retrieveConfig();`

packages/tempus-client_v3/src/config/ConfigManager.ts
lazarmitic: It should be, I forgot to make it private after doing some tests.

lazarmitic: fixed

packages/tempus-client_v3/src/config/ConfigManager.ts
lazarmitic: fixed

PR #1579, opened 2022-05-10
first integration with react-i18next

ShawTim:
- integrate with `react-i18next`
- update `i18next` locale in `useLocale()` when locate is updated
- auto defect browser locale and set it correspondingly by default
- store the locale preference in `localStorage` and `cookie`
- use `const { t } = useTranslation()` and `t('XXX.YYY')` for translations
- apply on `SettingsDropdown` and `SlippageInput`
- use `npm run i18next-parser` to extract all strings that needed to be translated
- use `npm run pseudo-translation` to fill pseudo translation
- added pseudo locale `zz`, will only available for development mode
- react provider `I18nProvider` for unit test to load `react-i18next` if necessary


packages/tempus-client_v3/src/components/SettingsDropdown/SettingsDropdown.tsx
```diff
@@ -1,20 +1,25 @@
...
-const SettingsDropdown: FC = () => (
-  <TooltipWrapper tooltipContent={<SettingsPopup />} placement="bottom-left">
-    <div className="tc__navbar-actions-dropdown">
-      <Typography variant="body-primary" color="text-primary-inverted">
-        Settings
-      </Typography>
-      <div className="tc__navbar-actions-dropdown-icon">
-        <Icon variant="down-chevron" size="tiny" color={colors.textPrimaryInverted} />
+const SettingsDropdown: FC = () => {
+  const { t } = useTranslation();
```
danieledamiani, line 9: shall we rename it to `translate`?

danieledamiani (approved):

packages/tempus-client_v3/src/components/SettingsDropdown/SettingsDropdown.tsx
ShawTim: the `t` here is provided by `useTranslation()` from `react-i18next`, unless we want to do:
```
const { t as translate } = useTranslation();
```
which is too bunky

PR #1584, opened 2022-05-12
Add chart dot

iamvukasin:


packages/tempus-client_v3/src/components/shared/Chart/Chart.tsx
```diff
@@ -75,6 +83,14 @@ function Chart<X, Y extends ValueType>(
     [],
   );
 
+  const chartDot = useCallback(
+    dotProps => {
+      const { payload, index, cx, cy } = dotProps;
+      return dot?.(payload.x, payload.y, index, cx, cy) ?? <></>;
```
ShawTim, line 89: is `<></>` equivalent to `null`?

iamvukasin: Yes, but the compiler doesn't allow `null` here

packages/tempus-client_v3/src/components/shared/Chart/DateChart.tsx
```diff
@@ -9,11 +9,18 @@ interface DateChartProps {
   yTick?: ChartTick;
   yTickFormatter?: (value: number, index: number) => string;
   topPercentageProjected?: number;
+  dot?: (
```
ShawTim, line 12: the type of `dot` is reused in several places, maybe we can consider to create a type for that

ShawTim (approved):

PR #1586, opened 2022-05-12
Silence unimportant warnings in some tests

iamvukasin:


packages/tempus-client_v3/src/components/shared/Chart/Chart.spec.tsx
```diff
@@ -11,6 +11,18 @@ const defaultProps: ChartProps & ChartSizeProps = {
 const subject = (props: ChartProps & ChartSizeProps) => render(<Chart {...props} />);
 
 describe('Chart', () => {
+  let originalConsoleWarn: (...consoleData: any[]) => void = () => {};
```
ShawTim, line 14: i think you can do the code here:
```
const originalConsoleWarn = console.warn;
```

ShawTim (approved):

PR #1594, opened 2022-05-13
Fix fixed APR calculation

lazarmitic:


packages/tempus-core-services/src/services/TempusPoolService.ts
```diff
@@ -292,6 +296,22 @@ export class TempusPoolService {
     throw new Error(`TempusPoolService - currentInterestRate() - Address '${address}' is not valid`);
   }
 
+  public async currentInterestRateStatic(address: string, overrides?: CallOverrides): Promise<BigNumber> {
+    const contract = this.tempusPoolsMap[address];
+    if (contract) {
+      try {
+        if (overrides) {
+          return await contract.callStatic.currentInterestRate(overrides);
```
yuval-weiss, line 304: should be `updateInterestRate`

packages/tempus-core-services/src/services/TempusPoolService.ts
```diff
@@ -292,6 +296,22 @@ export class TempusPoolService {
...
+        }
+        return await contract.callStatic.currentInterestRate();
```
yuval-weiss, line 306: should be `contract.callStatic.updateInterestRate`

packages/tempus-core-services/src/services/TempusPoolService.ts
lazarmitic: updateInterestRate is internal :(

packages/tempus-core-services/src/services/TempusPoolService.ts
lazarmitic: updateInterestRate is internal :(

lazarmitic:
Added event time for scale factor calculation instead of using Date.now().

packages/tempus-client_v2/src/adapters/PoolDataAdapter.ts
```diff
@@ -765,28 +766,55 @@ export default class PoolDataAdapter {
...
+      ]);
+      const allEvents = [...depositedEvents, ...swapEvents, ...redeemedEvents];
+      let latestEventBlockNumber = 0;
+      allEvents.forEach(event => {
+        if (event.blockNumber > latestEventBlockNumber) {
+          latestEventBlockNumber = event.blockNumber;
+        }
+      });
+      let currentFixedAPRTime: number;
+      // In case event time was found
```
yuval-weiss, line 784: when does that happen?

lazarmitic: Newly deployed pool that does not have any actions taken on it. (From user perspective this will never happen because we always initialize the pool before pushing it into production). But for us, it will happen right after we deploy pool and someone wants to open the app to provide initial liquidity.

packages/tempus-core-services/src/services/TempusPoolService.ts
yuval-weiss, line 304: there's no point in using `callStatic` since we ended up using `currentInterestRate` and not `updateInterestRate`.

packages/tempus-client_v2/src/adapters/PoolDataAdapter.ts
```diff
@@ -765,28 +766,55 @@ export default class PoolDataAdapter {
...
+      }
+      // In case there are no events for tempus pool, we fallback to Date.now() instead of latest event time
+      else {
+        currentFixedAPRTime = Date.now();
+      }
+
+      const poolTimeRemaining = (maturityDate - currentFixedAPRTime) / 1000;
+      const scaleFactor = ethers.utils.parseEther(((SECONDS_IN_A_DAY * DAYS_IN_A_YEAR) / poolTimeRemaining).toString());
 
       const principals = await this.statisticService.estimatedDepositAndFix(
```
yuval-weiss, line 797: Let's add `{ blockTag: latestEventBlock }`

PR #1599, opened 2022-05-16
Add useSelectedChain hook

lazarmitic:


packages/tempus-client_v3/src/hooks/useSelectedChain.ts
```diff
@@ -0,0 +1,6 @@
+import { bind } from '@react-rxjs/core';
+import { createSignal } from '@react-rxjs/utils';
+
+export const [selectedChain$, setSelectedChain] = createSignal<string | null>();
```
danieledamiani, line 4: do we need to export `selectedChain$`?

packages/tempus-client_v3/src/hooks/useSelectedChain.ts
```diff
@@ -0,0 +1,6 @@
+import { bind } from '@react-rxjs/core';
```
danieledamiani, line 1: we should add an unit test, even if it's trivial

ShawTim (approved):

packages/tempus-client_v3/src/hooks/useSelectedChain.ts
lazarmitic: Got it, I will read docs that you sent for stream testing and add tests :D

packages/tempus-client_v3/src/hooks/useSelectedChain.ts
lazarmitic: If we want to pipe other states to this one yes - for example, when chain changes, I want to update tokenBalances for example.

danieledamiani (approved):

PR #1613, opened 2022-05-18
Add preview step for deposit modal

iamvukasin:


ShawTim:
reviewing

packages/tempus-client_v3/src/components/CurrencyInputModal/CurrencyInputModal.tsx
```diff
@@ -81,62 +102,78 @@ const CurrencyInputModal: FC<CurrencyInputModalProps> = props => {
...
+              onClick={handlePreviewButtonClick}
+              variant="primary"
+              size="large"
+              fullWidth
+              state="default"
+            />
+          </div>
+        </>
+      )}
+      {content === 'action' && (
```
ShawTim, line 133: maybe we want to create a private component `PreviewModal` and `ActionModal` (whatever name) here to keep things in the top level clean, such that:
```
{content === 'preview' && <PreviewModal />}
{content === 'action' && <ActionModal />}
```
easier to follow the render logic. but on the other hand, we need to see how many `props` needed to be passed into them. for sure we dont want to have a massive props becoz if that's the case it doesnt make it "clean". so it depends on the case and we need to find a balance. just a suggestion but not a must.

packages/tempus-client_v3/src/components/DepositModal/DepositModal.stories.tsx
```diff
@@ -66,6 +66,7 @@ const Template: ComponentStory<typeof DepositModal> = () => {
       <DepositModal
         inputPrecision={18}
         usdRates={singleCurrencyUsdRates}
+        poolStartDate={new Date(2022, 3, 1)}
```
ShawTim, line 69: just a question, will the date in unit test be expired after half year or a year? in theory the pool should have been matured and the test will fail without doing anything?

packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
```diff
@@ -11,6 +11,7 @@ type ModalSize = 'small' | 'large';
 export interface ModalProps {
   open: boolean;
   onClose: () => void;
+  onBack?: () => void;
```
ShawTim, line 14: i'm thinking should we have `onBack` on the `Modal` or should we keep the `onBack` on the `children` or modal content.
becoz it doesnt make much senses to have "Back" on `Modal`.
it should be a `Modal`, with a content that can switch between 2 nav state

danieledamiani:
reviewing

ShawTim:
done review with some comments

packages/tempus-client_v3/src/components/DepositModal/DepositModal.stories.tsx
danieledamiani: Agree, better to have a dynamic start date

danieledamiani (commented):
Nothing to add beside @ShawTim comments

packages/tempus-client_v3/src/components/DepositModal/DepositModal.stories.tsx
iamvukasin: This line is from stories that have mocked dates, so it doesn't use data from real pools anyways.

For unit tests, I think we will still use mocked pools/dates so we can see how the component renders and behaves without relying on real data and current date.

packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
ShawTim: to resolve this i think we can accept `title` (or `header`?) as a `ReactNode | string`, then we can render whatever we want. in this PR case we can pass a header `ReactNode` which render the title and the back button with `onBack` handling.
this way the `Modal` will be more generic and free from business/navigation logic.

packages/tempus-client_v3/src/components/CurrencyInputModal/CurrencyInputModal.tsx
iamvukasin: I've separated the content into two components as you proposed. In the future, `ModalActionContent` won't need `balance` and `inputPrecision` as it will be able to get them from hooks.

packages/tempus-client_v3/src/components/shared/Modal/Modal.tsx
iamvukasin: Good idea

ShawTim (approved):

PR #1648, opened 2022-05-19
Fetch yield-bearing token price rates

iamvukasin:


danieledamiani:
in review

packages/tempus-client_v3/src/hooks/useTokenRates.spec.ts
```diff
@@ -1,4 +1,5 @@
 import { renderHook } from '@testing-library/react-hooks';
+import { BigNumber } from 'ethers';
```
danieledamiani, line 2: as mentioned, would be better to use `Decimal` only in this package

packages/tempus-client_v3/src/hooks/useTokenRates.ts
```diff
@@ -32,35 +46,76 @@ interface TokenRateMap {
...
         token,
       );
       return tokenRate.pipe(
-        map(
-          rate =>
-            ({
-              [`${chain}-${address}`]: rate,
-            } as TokenRateMap),
-        ),
+        map(rate => {
```
danieledamiani, line 109: Nested ternary operators may look cool but are hard to read. Please use a duller `if` block

danieledamiani (commented):
a comment

packages/tempus-core-services/src/services/TempusPoolService.ts
```diff
@@ -304,4 +335,16 @@ export class TempusPoolService {
...
+  private async getTokenPrecision(tokenAddress: string): Promise<number | null> {
+    if (!this.chain) {
+      return null;
+    }
+
+    const contract = new ERC20Contract(this.chain, tokenAddress);
+    return contract
+      .decimals()
+      .then(decimals => decimals)
+      .catch(() => DEFAULT_DECIMAL_PRECISION);
```
lazarmitic, line 348: It's not a good idea to silently return invalid token precision in case call fails - this will just mask the error and cause issues in the app because we will end up using wrong token precision.

packages/tempus-core-services/src/services/TempusPoolService.ts
```diff
@@ -304,4 +335,16 @@ export class TempusPoolService {
     }
     throw new Error(`TempusPoolService - getFeesConfig() - Address '${address}' is not valid`);
   }
+
+  private async getTokenPrecision(tokenAddress: string): Promise<number | null> {
```
lazarmitic, line 339: Maybe we can create a BaseService class that contains util functions like this, and all other service classes can extend it. (this will help us avoid code repetition).

Second option that comes to my mind is to create service-utils directory and store functions like this in it, outside of service classes.

packages/tempus-core-services/src/services/TempusPoolService.ts
```diff
@@ -279,11 +300,21 @@ export class TempusPoolService {
     throw new Error(`TempusPoolService - currentInterestRate() - Address '${address}' is not valid`);
   }
 
-  public async currentInterestRate(address: string): Promise<BigNumber> {
+  public async currentInterestRate(address: string): Promise<Decimal> {
     const contract = this.tempusPoolsMap[address];
     if (contract) {
       try {
-        return await contract.currentInterestRate();
+        const yieldBearingTokenPrecision = await this.getTokenPrecision(await contract.yieldBearingToken());
```
lazarmitic, line 307: I think it's better to use info about token address/precision from config because it's already loaded into the app.

We should create Contract class for TempusPool-v1.1 and add handle decimal conversion in it.

In addition to that contract class we need to create service class that will be used by hooks/components to fetch data. We can call this service class `InterestRatesService` - inside this class we can handle different contract versions etc.

TempusPool service is built to work only with v1.1 contracts, once backend team deploys v2.0 of contracts, TempusPoolService class will not work most likely - so we should start building out a structure that will easily support v2.0 of contracts.

lazarmitic (approved):

PR #1697, opened 2022-05-27
Convert `useFixedApr` to pool-APR map

iamvukasin:


packages/tempus-client_v3/src/hooks/useFixedApr.ts
```diff
@@ -1,112 +1,106 @@
...
+  tokenAmount: Decimal,
+  isBackingToken: boolean,
+): Observable<Decimal> =>
+  from(
+    (statisticsService as StatisticsService).estimatedDepositAndFix(
+      tempusAMMAddress,
+      tokenAmount.toBigNumber(),
+      isBackingToken,
+    ),
+  ).pipe(switchMap(value => of(new Decimal(value))));
```
ShawTim, line 42: i think we can just do `.pipe(map(value => new Decimal(value)))` here?

packages/tempus-client_v3/src/hooks/useFixedApr.ts
```diff
@@ -1,112 +1,106 @@
...
+  callOverrideData?: { blockTag: number },
+): Observable<Decimal> =>
+  from(
+    (statisticsService as StatisticsService).estimatedMintedShares(
+      tempusPoolAddress,
+      tokenAmount.toBigNumber(),
+      isBackingToken,
+      callOverrideData,
+    ),
+  ).pipe(switchMap(value => of(new Decimal(value))));
```
ShawTim, line 58: same as above

ShawTim:
if it returns a map which contains a bunch of data then i'd prefer to rename it as `useFixedAprs`.

ShawTim (approved):

PR #1704, opened 2022-05-30
Set branch to fetch config from in env file

lazarmitic:


ShawTim (approved):

danieledamiani:
in review

packages/tempus-client_v3/src/config/ConfigManager.ts
```diff
@@ -5,6 +5,7 @@ const TempusPoolsConfig = {
   owner: 'tempus-finance',
   repo: 'tempus-pools-config',
   path: 'config.json',
+  ref: process.env.REACT_APP_CONFIG_BRANCH,
```
danieledamiani, line 8: If `ref` is `undefined` does it revert to `master` or should it be explicited?

danieledamiani (approved):
one comment

packages/tempus-client_v3/src/config/ConfigManager.ts
lazarmitic: Yes, default branch is branch that is set as a default in target repo (in this case master).

PR #1713, opened 2022-05-31
Fix slippage input allowing values above 100%

iamvukasin:


packages/tempus-client_v3/src/components/shared/SlippageInput/SlippageInput.scss
```diff
@@ -7,7 +7,6 @@ $button-width: 54px;
     }
     .tc__base-input {
       border-radius: 4px 0 0 4px;
-      width: calc(100% - $button-width);
```
ShawTim, line 10: can you check and see whether it's safe to remove this on application on storybook, and try adjusting the width of the component and see?

packages/tempus-client_v3/src/components/shared/SlippageInput/SlippageInput.tsx
```diff
@@ -32,7 +32,13 @@ const SlippageInput: FC<SlippageInputProps> = props => {
   }, [onAutoUpdate]);
   const handleDebounceChange = useCallback(
     (inputValue: string) => {
-      onPercentageUpdate?.(Decimal.parse(inputValue, 0).div(100));
+      let resolvedValue = Decimal.parse(inputValue, 0).div(100);
```
ShawTim, line 35: think this can be:
```
const resolvedValue = Decimal.parse(inputValue, 0).div(100);
onPercentageUpdate?.(resolvedValue.gt(1) ? ONE : resolvedValue);
```

not a fan of `let` but it's personal preference only. but you may want to use `ONE` here

ShawTim (approved):

PR #1717, opened 2022-05-31
Improve `useConfig`

danieledamiani:


packages/tempus-client_v3/src/hooks/useConfig.ts
```diff
@@ -1,20 +1,27 @@
...
+
+interval(CONFIG_RETRYING_INTERVAL_IN_MS)
+  .pipe(
+    startWith(0),
+    map(() => getConfigManager().getConfig()),
+    filter(config => Object.keys(config).length > 0),
+    tap(config => configSubject$.next(config)),
+    take(1),
+  )
+  .subscribe();
```
ShawTim, line 18: why do we need to `subscribe()` at the end?
or, should we move `configSubject$.next(config)` to `subscribe()`?

danieledamiani: > why do we need to `subscribe()` at the end?

Because it's necessary to subscribe to an observable in order to receive the streamed data.

danieledamiani: > or, should we move configSubject$.next(config) to subscribe()?

it's equivalent

ShawTim (approved):

PR #1722, opened 2022-06-01
Update useChainList hook

lazarmitic:
Moves useChain hook in separate file. Gets chain list directly from config since it's local now.


danieledamiani:
in review

danieledamiani (approved):

PR #1729, opened 2022-06-02
Add loading state for pool card

iamvukasin:


danieledamiani:
in review

packages/tempus-client_v3/src/components/shared/PoolCard/PoolCard.tsx
```diff
@@ -52,16 +53,16 @@ const PoolCard: FC<PoolCardProps> = props => {
   const earliestTerm = useMemo(() => min(terms), [terms]);
 
   const maxApr = useMemo(() => {
-    let max = new Decimal(0);
+    let max: Decimal | undefined;
     aprValues.forEach(aprValue => {
-      if (aprValue.gte(max)) {
+      if (aprValue && aprValue.gte(max ?? ZERO)) {
```
danieledamiani, line 58: you can initialise `max = ZERO`

danieledamiani (approved):
lgtm

packages/tempus-client_v3/src/components/shared/PoolCard/PoolCard.tsx
iamvukasin: I intentionally want `max` to stay `undefined` if all `aprValues` are `undefined`, because of `const loading = !maxApr`. Initializing `max` to `ZERO` would break that.

danieledamiani: Ok, I see.

A better way would be for `useFixedAprs` to return something like 
```
export interface PoolFixedAprMap {
  [chainPoolAddress: string]: { loading: boolean; apr: Decimal; } 
}
```

and use `loading` as flag. But we can refactor it later

danieledamiani (approved):

PR #1731, opened 2022-06-02
Show "Insufficient balance" error in currency input modal when balance is zero

iamvukasin:


packages/tempus-core-services/src/datastructures/Decimal.ts
```diff
@@ -81,6 +81,12 @@ export default class Decimal {
     return new Decimal(this.value.abs());
   }
 
+  eq(another: Numberish): boolean {
```
danieledamiani, line 84: I think `comparable` may be a better name for the argument

packages/tempus-client_v3/src/i18n/translations/es.json
```diff
@@ -7,6 +7,7 @@
     "title": ""
   },
   "CurrencyInputModal": {
+    "insufficientBalanceError": "",
```
danieledamiani, line 10: `Saldo insuficiente`

danieledamiani (dismissed):

danieledamiani (approved):

PR #1734, opened 2022-06-03
Add useTokenBalance hook

lazarmitic:
Contains one BehaviorSubject per token, this hook will eventually replace useWalletBalances hook.


packages/tempus-client_v3/src/hooks/usePoolBalances.ts
```diff
@@ -14,7 +15,9 @@ interface PoolBalanceMap {
 export const poolBalances$: Observable<PoolBalanceMap> = combineLatest([
   poolList$,
   walletAddress$,
-  walletBalances$,
+  walletBalances$, // TODO - Instead of depending on all token balances (walletBalances$), this stream should
+  // update only when TPS, TYS or LP token balance changes for specific pool
+  servicesLoaded$,
```
ShawTim, line 20: need to check whether `servicesLoaded$` is `true` to go
or, we can change `servicesLoaded$` to emit event only when it's `true`

packages/tempus-client_v3/src/hooks/useTokenBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+  subject: BehaviorSubject<Decimal | null>;
+  address: string;
+  chain: Chain;
+}
+
+// Each token will have a separate BehaviorSubject, and this map will store them with some additional token data
+const tokenBalanceDataMap = new Map<TokenChainAddressId, TokenBalanceData>();
+
+// Create a BehaviorSubject for each token in the config file
+const tokenList = getConfigManager().getTokenList();
```
ShawTim, line 20: the only concern here is we will always assume `getConfigManager().getTokenList()` is available at declaration time.
which means `Config` will always be part of the build.
i'm fine with this but we need to understand that we cannot go back

packages/tempus-client_v3/src/hooks/useTokenBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+  // TODO - Include servicesLoaded$ stream here as well
+  mergeMap(([walletAddress]) => {
+    const tokenBalanceFetchMap = [...tokenBalanceDataMap.values()].map(tokenBalanceData => {
+      if (!walletAddress) {
+        return of(null);
+      }
+
+      const services = getServices(tokenBalanceData.chain);
+      if (!services) {
+        throw new Error('useWalletBalances - stream$ - Failed to get services');
```
ShawTim, line 42: if we will throw Error then we may need to have `catchError` in the `pipe`

packages/tempus-client_v3/src/hooks/useTokenBalance.ts
```diff
@@ -0,0 +1,105 @@
...
+import { BehaviorSubject, map, combineLatest, from, of, merge, mergeMap, Subscription, tap } from 'rxjs';
+import { Chain, Decimal, getServices } from 'tempus-core-services';
+import { getConfigManager } from '../config/getConfigManager';
+import { walletAddress$ } from './useWalletAddress';
+
+// Improves readability of the code
+type TokenChainAddressId = string;
+
+interface TokenBalanceData {
+  subject: BehaviorSubject<Decimal | null>;
```
ShawTim, line 11: `Subject` is an `Observable` object so let's use `subject$` here

packages/tempus-client_v3/src/hooks/useWalletBalances.ts
```diff
@@ -1,46 +1,46 @@
...
-      Promise.all(
-        result.tokenList.map(async token => {
-          const services = getServices(token.chain);
-          if (!services) {
-            throw new Error('walletBalances$ - Failed to get services');
-          }
-          const balance = await services.WalletBalanceService.getTokenBalance(token.address, result.walletAddress);
+// Every time wallet address changes, we want to fetch wallet balances
+export const walletBalances$ = combineLatest([walletAddress$, servicesLoaded$]).pipe(
+  filter(([walletAddress]) => Boolean(walletAddress)),
```
danieledamiani, line 10: `filter(([walletAddress, serviceLoaded]) => Boolean(walletAddress) && serviceLoaded),

packages/tempus-client_v3/src/hooks/useWalletAddress.ts
```diff
@@ -1,6 +1,15 @@
+import { BehaviorSubject } from 'rxjs';
 import { bind } from '@react-rxjs/core';
-import { createSignal } from '@react-rxjs/utils';
 
-export const [walletAddress$, setWalletAddress] = createSignal<string>();
+export const walletAddress$ = new BehaviorSubject<string>('');
 
-export const [useWalletAddress] = bind<string>(walletAddress$, '');
+export function setWalletAddress(address: string): void {
+  const currentAddress = walletAddress$.getValue();
```
danieledamiani, line 7:
```
of(address)
.pipe(
  combineLatestWith(walletAddress$),
 filter(([address, currentAddress]) => address !== currentAddress),
 tap(([address]) => walletAddress$.next(address))
)

File: packages/tempus-client_v3/src/hooks/usePoolBalances.ts [hunk 7]
```diff
@@ -14,7 +15,9 @@ interface PoolBalanceMap {
 export const poolBalances$: Observable<PoolBalanceMap> = combineLatest([
   poolList$,
   walletAddress$,
-  walletBalances$,
+  walletBalances$, // TODO - Instead of depending on all token balances (walletBalances$), this stream should
```
danieledamiani, line 18: we can filter out `walletBalances$[poolAddress]`

packages/tempus-client_v3/src/hooks/useTokenBalance.ts
danieledamiani: or encapsulate the code in a function that gets called once the config is loaded

packages/tempus-client_v3/src/hooks/useWalletAddress.ts
lazarmitic: I would need to subscribe to this stream to make it run, and then immediately unsubscribe to prevent memory leaks, it seems a bit hacky...

packages/tempus-client_v3/src/hooks/usePoolBalances.ts
lazarmitic: Yes, but the problem is we don't know if balance of the tokens for pool with `poolAddress` changed in last update or not.

danieledamiani:
in review

packages/tempus-client_v3/src/hooks/useWalletAddress.ts
danieledamiani:
```
of(address)
.pipe(
  combineLatestWith(walletAddress$),
  filter(([address, currentAddress]) => address !== currentAddress),
  tap(([address]) => walletAddress$.next(address)),
  take(1)
)
.subscribe()
```

lazarmitic: I didn't realize that take(1) automatically does not keep the subscription active, thanks!

danieledamiani (approved):

PR #1736, opened 2022-06-03
Show "Connect Wallet" button in deposit modal if wallet is not connected

iamvukasin:


packages/tempus-client_v3/src/components/Wallet/Wallet.spec.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const { getByRole } = subject({ redirectTo: '/pool/invalid/lido' });
+
+    const connectButton = getByRole('button');
+
+    fireEvent.click(connectButton);
+
+    await waitFor(() => {
+      expect(mockConnect).toBeCalledTimes(1);
+      expect(mockSetChain).not.toBeCalled();
+    });
```
danieledamiani, line 118: should we add `expect(window.location.pathname).toBe(???);` ?

danieledamiani (approved):
lgtm, only one comment

packages/tempus-client_v3/src/components/Wallet/Wallet.spec.tsx
```diff
@@ -0,0 +1,153 @@
+import { act, fireEvent, render, waitFor } from '@testing-library/react';
+import { useConnectWallet, useSetChain } from '@web3-onboard/react';
+import _ from 'lodash';
```
ShawTim, line 3: is it a wrong import? plus, i think we want to get rid of `lodash` since we dont need it

iamvukasin: It was. Removed

ShawTim (approved):

ShawTim (approved):

PR #1743, opened 2022-06-06
Show deposit modal as overlay

iamvukasin:
image


danieledamiani (dismissed):

ShawTim:
reviewing

packages/tempus-client_v3/src/components/App/App.tsx
```diff
@@ -68,8 +68,9 @@ const App = () => {
                 element={<PoolPositionModalResolver />}
               />
               <Route path="/withdraw/:chain/:ticker/:protocol/:poolAddress" element={<WithdrawModalResolver />} />
-              <Route path="/pool/:chain/:ticker/:protocol" element={<DepositModalResolver />} />
-              <Route path="/" element={<Markets />} />
+              <Route path="/" element={<Markets />}>
+                <Route path="pool/:chain/:ticker/:protocol" element={<DepositModalResolver />} />
```
ShawTim, line 72: why do we want to:
- move only this `Route` under the root path
- remove the leading `/` in the path?

packages/tempus-client_v3/src/components/Markets/Markets.tsx
```diff
@@ -7,6 +8,7 @@ const Markets: FC = () => (
   <div className="tc__app__markets">
     <MarketsSubheader />
     <MarketsPools />
+    <Outlet />
```
ShawTim, line 11: what's the need for `Outlet`?

packages/tempus-client_v3/src/components/App/App.tsx
iamvukasin: We want to have [nested routes](https://www.robinwieruch.de/react-router-nested-routes/) and "inject" `DepositModalResolver` into `Markets` as an overlay using [`Outlet`](https://github.com/tempus-finance/tempus-app/pull/1743/files#diff-f01d59df535ef31342af7ace46c04b7dfa7639c25d0b5255df0fee543236e003R11). For other paths, we don't want that behavior (for some we do, like `WithdrawModalResolver`, but waiting for other pages to be finished).

packages/tempus-client_v3/src/components/Markets/Markets.tsx
iamvukasin: Answered it on the first comment

ShawTim (dismissed):

danieledamiani (approved):

PR #1745, opened 2022-06-06
Add `useApproveToken` hook

danieledamiani:


packages/tempus-client_v3/src/hooks/useApproveToken.ts
```diff
@@ -0,0 +1,43 @@
...
+    const eRC20TokenServiceGetter = getServices(chain as Chain)?.eRC20TokenServiceGetter;
+    const Erc20TokenService = eRC20TokenServiceGetter?.(tokenAddress, chain, signer);
+
+    if (!Erc20TokenService) {
+      console.error('useApproveToken - Erc20TokenService not available');
+      return Promise.resolve(null);
+    }
+
+    try {
+      return Erc20TokenService.approve(spenderAddress, amount.toBigNumber());
```
lazarmitic, line 34: We need to pass number of decimal points toBigNumber() when converting Decimal to BigNumber. Let's make this mandatory parameter.

We should handle token decimal conversion inside contract/service classes.

lazarmitic: tempus-client_v3 should only use Decimals.

packages/tempus-client_v3/src/hooks/useApproveToken.ts
```diff
@@ -0,0 +1,43 @@
...
+    } catch (error) {
+      console.error('useApproveToken - Failed to approve token amount!', error);
+      return Promise.resolve(null);
+    }
+  }),
+  map(contractTransaction => (contractTransaction ? { success: true, contractTransaction } : { success: false })),
+);
+
+export const [useTokenApprove] = bind<ApproveTokenStatus>(tokenApproveStatus$, {});
```
lazarmitic, line 43: It would be nice if hook itself returns one more variable that holds status of approval. (for loading UI).

lazarmitic:
in review

ShawTim:
reviewing

packages/tempus-client_v3/src/hooks/useApproveToken.ts
```diff
@@ -0,0 +1,43 @@
...
+  amount: Decimal;
+  signer?: JsonRpcSigner;
+}
+
+export interface ApproveTokenStatus {
+  success?: boolean;
+  contractTransaction?: ContractTransaction;
+}
+
+export const [approveToken$, approveToken] = createSignal<ApproveToken>();
```
ShawTim, line 21: this way we can only handle one approval at one time for the entire application.
should this be an array?

packages/tempus-client_v3/src/hooks/useApproveToken.ts
ShawTim: i think we should wrap all the things in the hook (just like `useTranslation`)
maybe sth like:
```
const { approveToken, response, status } = useTokenApprove();
```

packages/tempus-core-services/src/services/getServices.ts
```diff
@@ -23,6 +24,7 @@ type ServiceMap = {
   VariableRateService: VariableRateService;
   StorageService: StorageService;
   WalletBalanceService: WalletBalanceService;
+  eRC20TokenServiceGetter: typeof getERC20TokenService;
```
iamvukasin, line 27: I prefer `erc20TokenServiceGetter` as it is easier for me to read

iamvukasin:
Reviewing

packages/tempus-client_v3/src/hooks/useApproveToken.ts
danieledamiani: refactored

packages/tempus-client_v3/src/hooks/useApproveToken.ts
danieledamiani: refactored

packages/tempus-client_v3/src/hooks/useApproveToken.ts
danieledamiani: I've thought to send back the original request so the component can filter out the relevant response

packages/tempus-core-services/src/services/getServices.ts
danieledamiani: renamed

lazarmitic (approved):

iamvukasin (approved):

PR #1751, opened 2022-06-07
Update deposit and withdraw modals' UI to match the latest design

iamvukasin:


packages/tempus-client_v3/src/i18n/translations/en.json
```diff
@@ -112,7 +112,7 @@
     "buttonUnsupported": "Unsupported"
   },
   "WithdrawModal": {
-    "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
+    "description": "Select the amount you you like to withdraw.",
```
danieledamiani, line 115: double `you`

packages/tempus-client_v3/src/i18n/translations/it.json
```diff
@@ -16,7 +16,7 @@
     "labelApproveDefault": "",
     "labelApproveLoading": "",
     "labelApproveSuccess": "",
-    "labelAvailableForDeposit": "",
+    "labelCurrentBalance": "",
```
danieledamiani, line 19: `Seleziona l'importo che vorresti ritirare`

packages/tempus-client_v3/src/i18n/translations/es.json
```diff
@@ -16,7 +16,7 @@
     "labelApproveDefault": "",
     "labelApproveLoading": "",
     "labelApproveSuccess": "",
-    "labelAvailableForDeposit": "",
+    "labelCurrentBalance": "",
```
danieledamiani, line 19: `Seleccione la cantidad que desea retirar`

danieledamiani (dismissed):
one comment

packages/tempus-client_v3/src/components/CurrencyInputModal/CurrencyInputModalInfoRow.tsx
```diff
@@ -8,34 +8,45 @@ export interface CurrencyInputModalInfoRowProps {
...
-          </div>
-        )}
-        <Typography variant="body-primary" type="mono" weight="medium">
-          {value}
-        </Typography>
-        <Typography variant="body-primary" weight="medium">
-          {currency}
-        </Typography>
-      </div>
+      {valueChange && (
```
ShawTim, line 22: just to confirm, can you check whether this will render nth when `valueChange` is falsy, or will it render `false`?

packages/tempus-client_v3/src/i18n/translations/it.json
iamvukasin: Is there a shorter form as the English version is "Current Balance"?

packages/tempus-client_v3/src/components/CurrencyInputModal/CurrencyInputModalInfoRow.tsx
iamvukasin: `valueChange` is either `'increase'`, `'decrease'` or `undefined`. It should never have `false` as a value.

Anyway, I've already checked it as we have two info rows in the deposit modal (one has `valueChange` set to `increase` and one doesn't).

packages/tempus-client_v3/src/i18n/translations/it.json
danieledamiani: Sorry, I've translated the wrong bit. It would be `Saldo corrente`

danieledamiani (approved):

PR #1756, opened 2022-06-08
handle error correctly for hooks that containing periodic polling

ShawTim:
handle errors for 3 hooks with periodic polling
tested with inserting random errors on core-services functions
the idea is to use `catchError` within `pipe` for the inner `Observable` (not the outer) becoz `catchError` will hijack the original `Observable` stream to a new one and thus the periodic polling stops at error coz it's replaced by the `Observable` emitted by `catchError`


danieledamiani:
in review

iamvukasin:
Reviewing

packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -32,20 +32,30 @@ const intervalBeat$: Observable<number> = interval(POLLING_INTERVAL_IN_MS).pipe(
 
 export const poolTvls$ = new BehaviorSubject<PoolTvlMap>(DEFAULT_VALUE);
 
+const getServicesNotNull = (chain: Chain) => {
```
danieledamiani, line 35: `getServicesByChain`?

packages/tempus-client_v3/src/hooks/useTvlData.ts
```diff
@@ -32,20 +32,30 @@ const intervalBeat$: Observable<number> = interval(POLLING_INTERVAL_IN_MS).pipe(
...
   const { chain, address, backingToken } = tempusPool;
-  const poolTvl$ = (getServices(chain as Chain)?.StatisticsService as StatisticsService).totalValueLockedUSD(
-    chain as Chain,
-    address,
-    backingToken,
-  );
-  return poolTvl$.pipe(
+
+  // handle all error in Observable.pipe()
+  return of({}).pipe(
```
danieledamiani, line 47: why not simply
```
of(getServicesNotNull(chain).StatisticsService.totalValueLockedUSD(chain, address, backingToken))).pipe(
map(
      tvl =>
        ({
          [`${chain}-${address}`]: tvl,
        } as PoolTvlMap),
    )
)
```

packages/tempus-client_v3/src/hooks/useTvlData.ts
ShawTim: the main differences between this function and the original one is this one ensure it's not null and return error. `getServicesByChain` looks like same as the original one

packages/tempus-client_v3/src/hooks/useFixedAprs.ts
```diff
@@ -47,46 +43,60 @@ const intervalBeat$: Observable<number> = interval(POLLING_INTERVAL_IN_MS).pipe(
...
 export const poolAprs$ = new BehaviorSubject<PoolFixedAprMap>(DEFAULT_VALUE);
 
-const fetchData = (tempusPool: TempusPool): Observable<PoolFixedAprMap> => {
-  const { address, poolId, chain, spotPrice, maturityDate } = tempusPool;
-
-  if (maturityDate < Date.now()) {
-    return of({
-      [`${chain}-${address}`]: ZERO,
-    } as PoolFixedAprMap);
+const getServicesNotNull = (chain: Chain) => {
```
danieledamiani, line 46: same as in `useTvlData`

packages/tempus-client_v3/src/hooks/useTvlData.ts
ShawTim: to catch the error, the functions that possible to throw error need to be within `pipe()`. error wont be caught if we put them in the first `of`

packages/tempus-client_v3/src/hooks/useTvlData.ts
iamvukasin: `getNonNullServices` could work, but I'm still not a big fan of having `null` in the name. I will try to come up with a better name.

ShawTim: that was my another pick actually. feel free to suggest a good name for this

iamvukasin: `getDefinedServices` as an alternative

iamvukasin: `getServiceOrThrow`

ShawTim: 3 pick 1 i will buy `getDefinedServices`

packages/tempus-client_v3/src/hooks/useTvlData.ts
danieledamiani: Then a `try..catch` may do. `getServicesNotNull` doesn't return an Observable so it can be managed outside the stream

ShawTim: updated. `catchError` to catch all possible errors from `pipe()`, `try-catch` to catch error otherwise

danieledamiani (dismissed):

PR #1768, opened 2022-06-10
refactor for state hooks

ShawTim:


packages/tempus-client_v3/src/hooks/useWalletAddress.ts
```diff
@@ -1,16 +1,13 @@
...
-export function setWalletAddress(address: string): void {
-  of(address)
-    .pipe(
-      combineLatestWith(walletAddress$),
-      filter(([newAddress, currentAddress]) => newAddress !== currentAddress),
-      tap(([newAddress]) => walletAddress$.next(newAddress)),
-      take(1),
-    )
-    .subscribe();
+export function useWalletAddress(): [string, (address: string) => void] {
```
lazarmitic, line 7: We need to skip state update if new address is same as current address.

The reason I changed this to a behavior subject is because react-rxjs resers the state when there are no subscribers left.

ShawTim: we can use `distinctUntilChanged()` for this behavior

danieledamiani:
in review

packages/tempus-client_v3/src/components/HookSubscriber/HookSubscriber.tsx
```diff
@@ -0,0 +1,25 @@
...
+  // to keep at least one subscriber of the stream insides the state hooks
+  useLocale();
+  useUserPreferences();
+  useSelectedChain();
+  usePoolViewOptions();
+  useServicesLoaded();
+  useWalletAddress();
+  useSigner();
+
+  // TODO: may consider to call stream$.subscribe() here instead of app start
```
danieledamiani, line 22: 👍

danieledamiani: I'd add an `useEffect` to subscribe (and unsubscribe)

danieledamiani (approved):

PR #1781, opened 2022-06-13
useAppEvent to emit and listen application events

ShawTim:
- `useAppEvent` for emits and listening events (`deposit` and `withdraw` right now)
- `useFixedAprs` and `useTvlData` are now the listeners (will be more) and fetch data when receiving events
- 100% test coverage
- TODO: need to `emitAppEvent` when we do deposit and withdrawal


lazarmitic:
in review

lazarmitic (approved):

PR #1784, opened 2022-06-14
Add useWithdraw hook

lazarmitic:


packages/tempus-client_v3/src/hooks/useWithdraw.ts
```diff
@@ -0,0 +1,106 @@
...
+      tokenBalance,
+      capitalsBalance,
+      yieldsBalance,
+      lpBalance,
+      slippage,
+      signer,
+    } = payload;
+
+    try {
+      const services = getServices(chain);
```
ShawTim, line 51: we can use `getDefinedServices` now and no longer need to check null (becoz it will throw the Error if it's null)
that means we need to catch the Error - which is already done here.
or you can leave it there. i'm working on refactoring hooks anyway

ShawTim (approved):

packages/tempus-client_v3/src/hooks/useWithdraw.ts
lazarmitic: Ah, didn't know this. I will change it to getDefinedServices.

danieledamiani (dismissed):

ShawTim (approved):

PR #1791, opened 2022-06-15
useAllowances hooks for fetching allowance

ShawTim:


packages/tempus-client_v3/src/hooks/useAllowances.ts
```diff
@@ -0,0 +1,131 @@
...
+          signer as JsonRpcSigner,
+        );
+        const spenderAddress = config[chain].tempusControllerContract;
+        return from(erc20TokenServiceGetter.getAllowance(walletAddress, spenderAddress)).pipe(
+          map(
+            bigNum =>
+              ({
+                [`${chain}-${address}`]: {
+                  alwaysApproved: false,
+                  amount: new Decimal(bigNum, DEFAULT_DECIMAL_PRECISION),
```
lazarmitic, line 82: Services should return decimal instance. let's avoid using BigNumber outside of tempus-core-services.

lazarmitic, line 82: Precision should be same as token precision.

ShawTim: fixed

ShawTim: fixed

lazarmitic (approved):

PR #1798, opened 2022-06-16
Fix app breaking when canceling wallet connection

### LINKED ISSUE #1796: Closing wallet selection modal breaks the app
**To Reproduce**
1. Go to Markets.
2. Click on a pool card.
3. Click on the "Connect Wallet" button.
4. When the wallet selection modal appears, click on the close button without selecting any wallet.

**Expected behavior**
The modal closes and returns to the deposit preview modal.

**Actual behavior**
The app breaks.

iamvukasin:
Closes #1796

### DIFF
--- a/packages/tempus-client_v3/src/components/Wallet/Wallet.tsx
+++ b/packages/tempus-client_v3/src/components/Wallet/Wallet.tsx
@@ -95,6 +95,10 @@ const Wallet: FC<WalletProps> = props => {
 
   /**
    * Redirects after wallet is connected, if needed.
+   *
+   * TODO: Revisit this behavior as it may not be intended for other use cases.
+   * Right now, it redirects after wallet change, but should not redirect when
+   * user changes wallets, for example.
    */
   useEffect(() => {
     if (wallet && redirectTo) {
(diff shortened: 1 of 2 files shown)


packages/tempus-client_v3/src/components/Wallet/Wallet.tsx
```diff
@@ -93,8 +93,11 @@ const Wallet: FC<WalletProps> = props => {
 
   const navigate = useNavigate();
 
-  const redirectAfterConnect = useCallback(() => {
-    if (redirectTo) {
+  /**
+   * Redirects after wallet is connected, if needed.
+   */
+  useEffect(() => {
+    if (wallet && redirectTo) {
```
ShawTim, line 100: what if the wallet changed here, say switching between wallet accounts. will it trigger a redirect here?

iamvukasin: Nothing would happen. Right now, we use `Wallet` with redirect in one place and it's shown only when a user doesn't have a connected wallet. It's not future-proof though.

iamvukasin: Nothing would happen. Right now, we use `Wallet` with redirect in one place and it's shown only when a user doesn't have a connected wallet. It's not future-proof though.

ShawTim: can you add a TODO so that we can revisit this part

ShawTim (dismissed):

ShawTim (approved):

PR #1806, opened 2022-06-20
Add success modal navigation logic

lazarmitic:


ShawTim (dismissed):

packages/tempus-client_v3/src/components/DepositModal/DepositModal.tsx
```diff
@@ -229,6 +231,18 @@ const DepositModal: FC<DepositModalProps> = props => {
...
+  const handleDepositInAnotherPoolClick = useCallback(() => {
+    navigate('/');
+  }, [navigate]);
+
+  const handleManagePortfolioClick = useCallback(() => {
+    navigate('/portfolio');
+  }, [navigate]);
+
+  const handleCloseSuccessModal = useCallback(() => {
+    navigate(-1);
```
iamvukasin, line 243: What if a user comes to the deposit modal directly with a link? I think it's better to use `navigate('/')` instead.

lazarmitic: But if user makes a deposit from portfolio page we want to navigate back to portfolio page. I was thinking to add url param `from` that will contain:
'markets' if user opened deposit from markets
'portfolio' if user opened deposit from portfolio
'' - if user opened direct link

using this 'from' param we can navigate back to correct page.

wdyt?

iamvukasin: Oh, I forgot that one can deposit from the portfolio page too. :+1: for `from` prop

ShawTim: i think we can check the navigation history provided by the hook, if there is a "previous page" then just do -1, else maybe just go back to root

lazarmitic: Since we don't have fully functional Portoflio page yet, this will be hard to test, I will just add `navigate('/')` as you suggested.

lazarmitic: @ShawTim we can do that as well 👍🏻 

Let's do this once we have a portfolio page to test with.

packages/tempus-client_v3/src/components/DepositModal/DepositModal.tsx
```diff
@@ -240,7 +240,7 @@ const DepositModal: FC<DepositModalProps> = props => {
   }, [navigate]);
 
   const handleCloseSuccessModal = useCallback(() => {
-    navigate(-1);
+    navigate('/');
```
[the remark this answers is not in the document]
lazarmitic: Yup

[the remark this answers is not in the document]
lazarmitic: done

iamvukasin (approved):

PR #1820, opened 2022-06-20
Fix link tabs not showing selected tab with nested routes

iamvukasin:


packages/tempus-client_v3/src/components/App/App.tsx
```diff
@@ -21,8 +21,16 @@ const App = () => {
   const { t } = useTranslation();
 
   const navigationLinks: PageNavigationLink[] = [
-    { text: t('App.navMarkets'), path: '/' },
-    { text: t('App.navPortfolio'), path: '/portfolio/overview' },
+    {
+      text: t('App.navMarkets'),
+      path: '/',
+      pathPatterns: ['/', '/pool/*'],
```
ShawTim, line 27: so if we have `pathPatterns`, we wont use `path`?

packages/tempus-client_v3/src/components/shared/Tabs/Tabs.tsx
```diff
@@ -15,23 +15,34 @@ export interface TabsProps {
...
-  const values = Children.map(children, child => child?.props.href ?? child?.props.value);
-  const selectedTabIndex = useMemo(
-    () => (values ? values.findIndex(tabValue => tabValue === value) : 0),
-    [value, values],
-  );
 
-  const tabWidth = useMemo(() => 100 / (values ? values.length : 1), [values]);
+  const selectedTabIndex = useMemo(() => {
+    const selectedIndex = Children.map(children, child => {
+      if (child?.props.hrefPatterns) {
```
ShawTim, line 21: move the question to here.
so apparently if the `props` contain `hrefPatterns`, `href` will not be used. am i correct?

iamvukasin: Yes. In that case, `hrefPatterns` has priority when checking `location.pathname` match. `href` would still be used inside `Tab` for links.

iamvukasin: Not having `hrefPatterns` effectively means that you have only match when `location.pathname === href`.

ShawTim: in that case then it actually not necessary to pass `href` (which is `path` in `App.tsx`) as param if `pathPatterns` is provided

ShawTim: or, is it possible to accept a `paths` params, which is `string | (string | Regexp)[]`?

iamvukasin: `path` is for link itself, but `pathPatterns` is used to check if the current URL pathname matches any of `pathPatterns`. We have to keep them both.

ShawTim (approved):

PR #1831, opened 2022-06-21
Add withdrawn amount in withdraw success modal

lazarmitic:


packages/tempus-core-services/src/utils/getWithdrawnAmountFromTx.ts
```diff
@@ -0,0 +1,31 @@
+import { ContractTransaction, ethers } from 'ethers';
+import { Decimal } from '../datastructures';
+import ERC20ABI from '../abi/ERC20.json';
+
+export async function getWithdrawnAmountFromTx(
```
iamvukasin, line 5: The function is very similar to `getDepositAmountFromTx`. Is there a way to have one instead?

lazarmitic: Yes, it's possible to create one function to handle both cases, but down the line we need to read receipt data for other actions as well, and I think if we try to create single function for all actions, this function will become unreadable monster.

iamvukasin (approved):

packages/tempus-client_v3/src/components/WithdrawModal/WithdrawModal.tsx
```diff
@@ -258,7 +268,7 @@ export const WithdrawModal: FC<WithdrawModalProps> = props => {
       )}
       {/* Show success modal if withdraw is finalized */}
       <SuccessModal
-        description={`You have withdrawn [AMOUNT] ${currency.ticker} with ${new Date(
+        description={`You have withdrawn ${withdrawnAmountFormatted} ${currency.ticker} with ${new Date(
```
ShawTim, line 271: we need to extract this as translation like what we did for `DepositModal`

lazarmitic: fixed

ShawTim (dismissed):

ShawTim (approved):

PR #1832, opened 2022-06-21
Show USD balance in portfolio overview page

iamvukasin:


packages/tempus-client_v3/src/components/Portfolio/PortfolioOverview/PortfolioInfoBox.tsx
```diff
@@ -17,17 +17,22 @@ const PortfolioInfoBox: FC<PortfolioInfoBoxProps> = props => {
...
-        </Typography>
-        <Typography
-          className="tc__app__portfolio-info-box-currency"
-          variant="subheader"
-          weight="medium"
-          color="text-secondary"
-        >
-          USD
-        </Typography>
+        {value !== null && (
```
ShawTim, line 20: what'e the difference between `null` and empty string here?

ShawTim: the component intends to accept `null` instead of empty string but i dont see why. if `null` is same as empty string here then we dont need to accept `null`

packages/tempus-client_v3/src/hooks/useUserUsdBalance.ts
```diff
@@ -0,0 +1,15 @@
+import { bind } from '@react-rxjs/core';
+import { map } from 'rxjs';
+import { Decimal, ZERO } from 'tempus-core-services';
+import { poolBalances$ } from './usePoolBalance';
+
+const userUsdBalance$ = poolBalances$.pipe(
```
ShawTim, line 6: do we need a new hook for this? thinking it can be part of the `usePoolBalance`. it's just a USD-sum of `usePoolBalance`

packages/tempus-client_v3/src/hooks/usePoolBalance.ts
```diff
@@ -105,22 +109,25 @@ poolList.forEach(tempusPool => {
             'yield-bearing',
           ),
         );
+        const balanceInUsdFetch$ = from(
```
lazarmitic, line 112: No need to fetch USD balance. We already have balance in backing token, and token rates hook, you can use those two to get USD amount in the component.

Let's try to avoid sending too many RPC calls.

iamvukasin: Good point! Will refactor based on it

packages/tempus-client_v3/src/hooks/useUserUsdBalance.ts
iamvukasin: Removed

packages/tempus-client_v3/src/hooks/usePoolBalance.ts
```diff
@@ -106,21 +109,23 @@ poolList.forEach(tempusPool => {
           ),
         );
 
-        return combineLatest([balanceInBackingTokenFetch$, balanceInYieldBearingTokenFetch$]);
+        return combineLatest([balanceInBackingTokenFetch$, balanceInYieldBearingTokenFetch$, tokenRates$]);
```
lazarmitic, line 112: Is this stream going to re-run when token rates stream is updated?

iamvukasin: I ran some tests that generate random token rates and it looked like it always re-ran after the update.

PR #1848, opened 2022-06-22
Show empty charts in portfolio overview if user doesn't have any deposits

iamvukasin:


ShawTim (dismissed):

packages/tempus-client_v3/src/components/Portfolio/PortfolioOverview/PortfolioValueChart.tsx
```diff
@@ -1,31 +1,38 @@
...
 interface DataPoint extends ChartDataPoint<Date, number> {
-  transactionType: TransactionType;
+  transactionType?: TransactionType;
 }
 
-const PortfolioValueChart: FC<ChartSizeProps> = props => {
+const PortfolioValueChart: FC<ChartCommonProps> = props => {
   const { width, height = 512 } = props;
+  const userDepositedPools = useUserDepositedPools();
+  const hideData = !userDepositedPools.length;
```
lazarmitic, line 17: Current user balance might be zero, but user might have had positions in the past. Do we want to show chart data in case current user value is zero, but user had a position previously?

iamvukasin: I think we should show the chart in that case because that user has a history of using our app and should be able to track it.

lazarmitic: In that case we should probably hide chart data using historical data (from graph) instead of only using current user position in the pool.

iamvukasin: I agree, ser

iamvukasin: Added TODOs so we can revisit it after we integrate the Graph.

PR #1889, opened 2022-06-23
Hide balance sorting if wallet is not connected

iamvukasin:


packages/tempus-client_v3/src/components/Markets/MarketsSubheader/MarketsSubheader.tsx
```diff
@@ -95,7 +97,7 @@ const MarketsSubheader = () => {
           <DropdownSelectableItem label={t('MarketsSubheader.sortMaturity')} value="maturity" />
           <DropdownSelectableItem label={t('MarketsSubheader.sortTVL')} value="tvl" />
           <DropdownSelectableItem label={t('MarketsSubheader.sortAPR')} value="apr" />
-          <DropdownSelectableItem label={t('MarketsSubheader.sortBalance')} value="balance" />
+          {walletAddress ? <DropdownSelectableItem label={t('MarketsSubheader.sortBalance')} value="balance" /> : null}
```
lazarmitic, line 100: does it work if you do 

`walletAddress && <DropdownSelectableItem ... />` ?

in order to avoid adding null to prop type.

lazarmitic: ah, let's leave it like this then.

PR #1906, opened 2022-06-27
Show real fees in `FeeTooltip`

iamvukasin:


packages/tempus-client_v3/src/hooks/useFees.ts
```diff
@@ -0,0 +1,52 @@
+import { bind } from '@react-rxjs/core';
```
lazarmitic, line 1: Let's check if fees are static (do not change over the lifespan of the pool), if that's the case we can put fees in config and just read them from config.

packages/tempus-client_v3/src/hooks/useFees.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const [poolForFees$, setPoolForFees] = createSignal<TempusPool>();
+
+const poolFees$ = (chain: Chain, tempusPoolAddress: string): Observable<Decimal[]> =>
+  getDefinedServices(chain).TempusPoolService.getFeesConfig(tempusPoolAddress);
+
+const swapFee$ = (chain: Chain, tempusAmmAddress: string): Observable<Decimal> => {
+  const { getChainConfig } = getConfigManager();
+  return from(getTempusAMMService(chain, getChainConfig).getSwapFeePercentage(tempusAmmAddress)).pipe(
+    map(fee => new Decimal(fee, DEFAULT_TOKEN_PRECISION)),
```
lazarmitic, line 29: Are we sure that fees are always 18 decimals?

Also, let's update services to return decimal instead of bignumber.

packages/tempus-client_v3/src/hooks/useFees.ts
```diff
@@ -0,0 +1,52 @@
...
+  );
+};
+
+const fees$ = poolForFees$.pipe(
```
lazarmitic, line 33: Instead of using poolForFees$ to set pool for which we want to fetch the fees, can we fetch fees for all pools once, store them in a subject as a map, and just read values from the map when needed. This way, we are going to re-fetch fees every time component that is showing fees is re-rendered.

lazarmitic: Also, do we need fees for mature pools? If not we can skip fetching fees data for mature pools.

packages/tempus-client_v3/src/hooks/useFees.ts
iamvukasin: It is changeable

packages/tempus-client_v3/src/hooks/useFees.ts
iamvukasin: It looks like [they are](https://github.com/tempus-finance/tempus-protocol/blob/88d4fad6ad3c6106eeb9f50030f1abc8665ba46d/contracts/amm/TempusAMM.sol#L47).

I skipped converting to `Decimal` here on purpose as it affects another function, so the PR would have been bigger. Can update it though.

lazarmitic: Let's update it in a separate PR :D

packages/tempus-client_v3/src/hooks/useFees.ts
lazarmitic: Makes sense, let's keep it because of withdrawal fee :D

danieledamiani (dismissed):

danieledamiani (dismissed):

ShawTim:
reviewing

packages/tempus-core-services/src/services/TempusPoolService.ts
```diff
@@ -340,14 +341,24 @@ export class TempusPoolService {
...
+                ? [depositPercent, earlyRedeemPercent, matureRedeemPercent]
+                : fees;
+
+            return percentArray.map(percent => new Decimal(percent, DEFAULT_DECIMAL_PRECISION));
+          }),
+        );
       } catch (error) {
         console.error('TempusPoolService - getFeesConfig() - Failed to fetch pool fees.', error);
-        return Promise.reject(error);
+        return throwError(() => new Error(error));
```
ShawTim, line 361: is it necessary to have `new Error(error)` instead of `error`? (becoz type is `unknown`?)

packages/tempus-client_v3/src/components/CurrencyInputModal/ModalActionContent.tsx
```diff
@@ -15,6 +16,7 @@ import {
 } from '../shared';
 
 interface ModalActionContentProps {
+  selectedPool?: TempusPool;
```
ShawTim, line 19: it looks like this is not an optional param?

ShawTim (approved):

packages/tempus-client_v3/src/components/CurrencyInputModal/ModalActionContent.tsx
iamvukasin: I'd keep it optional as it is only used for fee tooltip. If I change to `selectedPool: TempusPool`, I'd have to hide the modal content until the selected pool is loaded. The content should work without `selectedPool`.

PR #1907, opened 2022-06-27
Add useNegativePoolInterestRate hook

lazarmitic:


danieledamiani:
in review

packages/tempus-core-services/src/services/PoolInterestRateService.ts
```diff
@@ -0,0 +1,24 @@
...
+
+  async isPoolInterestRateNegative(poolAddress: string): Promise<boolean> {
+    const backingTokenPrecision = this.getPoolConfig(poolAddress).tokenPrecision.backingToken;
+
+    const tempusPoolContract = new TempusPoolV1Contract(this.chain, poolAddress);
+
+    const initialInterestRate = await tempusPoolContract.initialInterestRate(backingTokenPrecision);
+    const currentInterestRate = await tempusPoolContract.currentInterestRate(backingTokenPrecision);
+
+    return initialInterestRate.gt(currentInterestRate);
```
danieledamiani, line 22: is it ok to be `gt` and not `gte`?

danieledamiani (dismissed):

packages/tempus-core-services/src/services/PoolInterestRateService.ts
lazarmitic: Yes, if difference is 0 we should not mark interest rate as negative.

danieledamiani (approved):

PR #1951, opened 2022-06-29
Add FundsAvailable section

lazarmitic:


packages/website_v2/src/components/FundsAvailable/FundsAvailable.tsx
```diff
@@ -0,0 +1,41 @@
+import { useEffect, useMemo, useState } from 'react';
+import { Decimal, DecimalUtils } from 'tempus-core-services';
+import TreasuryValueService from '../../services/TreasuryValueService';
+
+import './FundsAvailable.scss';
+
+const FundsAvailable = (): JSX.Element => {
```
ShawTim, line 7: just a minor comment, i think "Funds Available" is the presentation naming, it's better to be "Treasury" or sth similar which is the actual meaning

ShawTim (approved):

packages/website_v2/src/components/FundsAvailable/FundsAvailable.tsx
lazarmitic: I just used name that was specified in the issue - I can change it :D

PR #1955, opened 2022-06-29
Add hover state for avatars in website

iamvukasin:


packages/website_v2/src/components/Hero/Hero.tsx
```diff
@@ -18,12 +18,54 @@ const Hero = (): JSX.Element => {
       </button>
       <div className="tw__hero-banner">
         <div className="tw__hero-banner-stripes" />
-        <ContributorAvatar name="avatar-1" x={8} y={7} />
-        <ContributorAvatar name="avatar-2" x={14} y={216} />
-        <ContributorAvatar name="avatar-3" x={25} y={28} />
-        <ContributorAvatar name="avatar-1" x={37} y={79} />
-        <ContributorAvatar name="avatar-2" x={73} y={166} />
-        <ContributorAvatar name="avatar-3" x={84} y={411} />
+        <ContributorAvatar
```
danieledamiani, line 21: we should put the avatar info in an array and then do
```
avatarArray.map(avatar => <ContributorAvatar avatarName={avatar.name} ...)
```

danieledamiani: so later will be easier to just update the info file instead of reworking the component

danieledamiani (approved):

PR #1958, opened 2022-06-30
Update footer section links layout

lazarmitic:


ShawTim (approved):

packages/website_v2/src/components/Footer/Footer.tsx
```diff
@@ -1,36 +1,83 @@
 import './Footer.scss';
 
-// TODO
-// contact email?
 const Footer = (): JSX.Element => (
   <div className="tw__footer">
+    <div className="tw__footer__background" />
```
danieledamiani, line 5: do we have the logo twice?

PR #1966, opened 2022-06-30
Add `Who are we` Section

danieledamiani:


packages/website_v2/src/index.scss
```diff
@@ -32,9 +32,30 @@ body {
...
+
+.tw__section__subsection-title {
+  font-style: normal;
+  font-weight: 600;
+  font-size: 28px;
+  line-height: 30px;
+  color: #EFECFF;
+}
+
+.affix {
```
iamvukasin, line 58: Is this from some library?

iamvukasin (approved):

packages/website_v2/src/index.scss
danieledamiani: going to remove it

PR #1998, opened 2022-07-01
Make `Who are we` section scrollable

iamvukasin:


packages/website_v2/src/components/Who/Who.tsx
```diff
@@ -26,47 +28,54 @@ They break the language barriers between Web3 and the real world.`.trim();
...
+    [activeSection],
+  );
+
+  const onScroll = useCallback(() => {
+    const container = document.querySelector('.tw__who') as HTMLDivElement;
+    const content = document.querySelector('.tw__who__content') as HTMLDivElement;
+    const scrollableHeight = container.clientHeight - content.clientHeight;
+    const percentageScrolled = content.offsetTop / scrollableHeight;
+    const selectedSection = SECTIONS[Math.min(Math.floor(percentageScrolled * SECTIONS.length), SECTIONS.length - 1)];
+    setActiveSection(selectedSection);
```
danieledamiani, line 58: We're invoking `setActiveSection` potentially multiple times per second

packages/website_v2/src/components/Who/Who.tsx
```diff
@@ -26,47 +28,54 @@ They break the language barriers between Web3 and the real world.`.trim();
...
   }, []);
 
+  useEffect(() => {
+    window.addEventListener('scroll', onScroll);
```
danieledamiani, line 62: we should also `removeEventListener` to avoid memory leaks

packages/website_v2/src/components/Who/Who.tsx
```diff
@@ -26,47 +28,54 @@ They break the language barriers between Web3 and the real world.`.trim();
...
-    setActiveSection(section as Sections);
+  const handleTitleClick = useCallback(
+    (event: MouseEvent<HTMLDivElement>) => {
+      const section = (event.currentTarget as HTMLDivElement).getAttribute('data-test') as Section;
+
+      if (section === activeSection) {
+        return;
+      }
+
+      const container = document.querySelector('.tw__who') as HTMLDivElement;
```
danieledamiani, line 41: some logic seems duplicated below

PR #2014, opened 2022-07-04
skeleton of team page

ShawTim:


lazarmitic (approved):

packages/website_v2/src/components/TeamPage/TeamPage.scss
```diff
@@ -0,0 +1,85 @@
+.tw__team {
+  .tw__team__header {
+    background: url(./header-background.png);
```
iamvukasin, line 3: Maybe we should use the SVG version?

PR #2026, opened 2022-07-05
Make header responsive

iamvukasin:


packages/website_v2/src/index.scss
```diff
@@ -24,6 +24,17 @@ body {
   font-family: articulat-cf, sans-serif;
 }
 
+.tw__btn {
+  padding: 0;
+  border: none;
+  background-color: transparent;
+  cursor: pointer;
+
+  &:disabled {
```
danieledamiani, line 33: we need some `hover` state

lazarmitic (approved):

packages/website_v2/src/index.scss
iamvukasin: This is just an unstyled button like we have `ButtonWrapper` for client v3. Will add TODO for menu and close buttons.

PR #2062, opened 2022-07-06
Improve transitions

danieledamiani:


packages/website_v2/src/components/shared/LinkBlock/LinkBlock.scss
```diff
@@ -12,7 +12,7 @@
   box-sizing: border-box;
   color: #050a4a;
   text-decoration: none;
-  transition: transform 0.3s ease-out;
+  transition: all 0.3s ease-out;
```
iamvukasin, line 15: There're extra spaces here

iamvukasin (approved):

packages/website_v2/src/components/shared/LinkBlock/LinkBlock.scss
danieledamiani: fixed

PR #2245, opened 2022-07-25
Convert `useFees`  to use `Observable<Decimal>`

iamvukasin:


packages/tempus-core-services/src/services/TempusAMMService.ts
```diff
@@ -149,14 +151,14 @@ export class TempusAMMService {
     );
   }
 
-  async getSwapFeePercentage(address: string): Promise<BigNumber> {
+  getSwapFeePercentage(address: string): Observable<Decimal> {
     const contract = this.tempusAMMMap.get(address);
     if (contract) {
       try {
-        return await contract.getSwapFeePercentage();
+        return from(contract.getSwapFeePercentage()).pipe(map(value => new Decimal(value, DEFAULT_DECIMAL_PRECISION)));
```
ShawTim, line 158: if `contract.getSwapFeePercentage()` throw an Error (or return an `Observable<Error>` or `Promise.reject()`, will it be caught by `try-catch`? or we will need to handle the error in the `pipe` through `catchError()` as well

ShawTim (approved):

packages/tempus-core-services/src/services/TempusAMMService.ts
iamvukasin: I think we need `catchError` as `getSwapFeePercentage` returns `throwError`.

iamvukasin: `contract.getSwapFeePercentage()` returns `Promise<BigNumber>`, so I guess it will be caught by `try-catch`. I will add some tests so we can have it covered.

danieledamiani (approved):

PR #2258, opened 2022-07-29
Remove language tests

irinelT:
Removed language tests because language switching button was removed.


ShawTim:
can we do a skip test instead of deleting it? we will need it later

ShawTim (approved):

PR #2271, opened 2022-08-03
E2e first deposit test

irinelT:
- [x] Added slippage to homepage tests.
- [x] Deposit tests with metamask approval on ETH fork network.


integration-testing/modules/manage.ts
```diff
@@ -16,9 +18,17 @@ export async function manageDepositSwitchCurrency(browser: BrowserContext, backi
...
+export async function manageDepositFillAmount(browser: BrowserContext, backingToken = 'USDC', currency = 'USDC', amount = '1')
 : Promise<Page> {
   const page: Page = await manageDepositSwitchCurrency(browser, backingToken, currency);
   await page.fill('.tc__currency-input__amount > input', amount);
   return page;
 }
+
+export async function deposit(browser: BrowserContext, backingToken = 'USDC', currency = 'USDC', amount = '0.1'): Promise<Page> {
+  const page: Page = await manageDepositFillAmount(browser, backingToken, currency, amount);
+  await page.click('text="Execute"');
```
danieledamiani, line 30: if we ever change the button label, this test will fail. Is there a way to target the button by id, class or some metatag?

integration-testing/tests/deposit.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+test.describe.serial('Deposit tests', () => {
+  let browser: BrowserContext;
+  test.beforeAll(async () => {
+    browser = await chromiumPersistant();
+    await metamaskLogin(browser);
+    await tempusMetamaskConnect(browser);
+    await tempusNetworkChange(browser, TEMPUS_ETH_NETWORK.name);
+  });
+
+  test('Deposit 1', async () => {
```
danieledamiani, line 18: should we specify `Deposit 1 ETH`?

integration-testing/modules/manage.ts
irinelT: Sure, its possible. Text is my favorite pick as it is descriptive and makes the code readable. Also, if you change the label I prefer to be notified and check if it is as designed. So I think text is the best selector but it is open for discussion.

danieledamiani (approved):

PR #2276, opened 2022-08-05
updated logos for website

ShawTim:


packages/website_v2/src/icons/ArrowRight.tsx
```diff
@@ -1,6 +1,8 @@
+import { memo } from 'react';
+
 const ArrowRight = (): JSX.Element => (
   <svg
-    className="tw__arrow-right-icon"
+    className="nw__arrow-right-icon"
```
danieledamiani, line 5: should be `tw__arrow`

danieledamiani (approved):

packages/website_v2/src/icons/ArrowRight.tsx
ShawTim: fixed

PR #2299, opened 2022-09-09
Website 2.1.1

danieledamiani:


.circleci/config.yml
```diff
@@ -114,59 +114,6 @@ jobs:
...
-            ls /root/project/test-results
-      - persist_to_workspace:
-          root: ./
-          paths:
-            - ./
-      - store_test_results:
-          path: /root/project/test-results
-      - store_artifacts:
-          path: /root/project/test-results
-
```
irinelT, line 169: The latest `./circle/config.yml` is very important. Please use the latest one and change only what is necessary.

irinelT (changes requested):

.circleci/config.yml
danieledamiani: restored

irinelT (approved):

PR #2305, opened 2022-09-13
Add historical fixed apr data in pool details modal

lazarmitic:


packages/tempus-client_v3/src/hooks/useFixedRatesData.ts
```diff
@@ -0,0 +1,67 @@
...
+
+      return merge(fixedRatesDataFetch$);
+    }
+    return of(null);
+  }),
+);
+
+export const [useFixedRatesData] = bind<FixedRateData[] | null>(fixedRatesData$, []);
+
+let streamSubscription: Subscription = fixedRatesDataStream$.subscribe();
```
ShawTim, line 56: we may want to subscribe the stream in `HookSubscriber` instead of JS loaded

ShawTim (approved):

